PR #75, opened 2015-02-05
Example `pysam.AlignedSegment` needs fixing under `v0.8.1`

PeteHaitch:
The example in the docs of creating a `pysam.AlignedSegment` no longer works under `v0.8.1`.This is because the `query_qualities` can no longer be set as a string.

I'm not sure what the preferred approach is for setting the `query_qualities`. I am using `pysam.fromQualityString()` but it could equally be hand-coded using `array.array()` with the type set to `B`.


AndreasHeger:
Thanks!

PR #76, opened 2015-02-09
Upgrade to htslib 1.2.1 and add support for htslib VCF/BCF API

bioinformed:
This code is incomplete and preliminary.  It does offer a nearly complete immutable Pythonic interface to VCF/BCF metadata and data with reading and writing capability, but has no capability (yet) to mutate the resulting data (beyond dropping all samples).  Documentation and unit test suite are still quite preliminary and will be improved quickly.  The code is also superficially specific to Python 2 and will require a bit of work to properly adapt to Python 3.  

The API and implementation should be considered alpha quality and will almost certainly continue to evolve without any claims of maintaining backward compatibility for the next few releases.


AndreasHeger:
Thanks! I will start merging it into pysam 0.8.2.

AndreasHeger:
@bioinformed , I have merged the changes into the 0.8.2 branch. Please let me know if you want to have a look before the next release.

Not sure how I mark this branch as merged...

bioinformed:
Looks great!  I've tested the code and that bdist_wheel works under both Linux and OS X.

AndreasHeger:
Thanks! I will release shortly.

PR #85, opened 2015-03-09
Updates docs to reflect htslib 1.2.1

jrandall:
The current version of pysam (0.8.2) has already been updated to htslib 1.2.1, but the docs still say it wrap htslib 1.1. This updates the dos to reflect that reality.


AndreasHeger:
Thanks!

PR #89, opened 2015-03-18
added filenames in error messages of tabix_compress.

TyberiusPrime:
Just correcting an oversight - the '%s' was already in the message, it just wasn't getting filled.


AndreasHeger:
Thanks!

Issue #88, opened 2015-03-17
Feature request: allow setting of htslib verbosity level


dcolligan:
There doesn't appear to be any way to set the verbosity level (`hts_verbose`) of htslib from pysam.  It would be useful to be able to control this value so one could e.g. ensure htslib didn't output anything while using pysam.

jeromekelleher:
This functionality was added in #90 and #91.

PR #90, opened 2015-03-19
Straw-man implementation of hts_verbose access.

jeromekelleher:
This is a straw-man partial implementation of providing access to `hts_verbose`, so we can control when messages are written to the low-level file descriptors. (See issue #88).

Problems with this PR:

1. It doesn't work. I haven't been able to figure out how to get the linkage for the extern hts_verbose variable to work. It remains set to 3 no matter what I do. If this is a genuine problem (and not just my Cython idiocy), perhaps we should open an issue with htslib where they provide functions to get and set the verbosity instead of requiring us to use a global variable?
2. This is in the wrong file. I think this functionality probably belongs in `chtslib.pyx`, but I couldn't get the compiler to see `hts_verbose` if  I put it in there. 

I'm new to Cython, so any pointers would be much appreciated!

### DIFF
--- a/pysam/cbcf.pyx
+++ b/pysam/cbcf.pyx
@@ -196,7 +196,19 @@ from cpython cimport PyBytes_Check, PyUnicode_Check
 from cpython.version cimport PY_MAJOR_VERSION
 
 
-__all__ = ['VariantFile', 'VariantHeader']
+__all__ = ['VariantFile', 'VariantHeader', 'set_hts_verbose', 'get_hts_verbose']
+
+
+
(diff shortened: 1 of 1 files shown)


AndreasHeger:
Hi @jeromekelleher , the approach I usually take is to write getter/setters in C and call them from cython.
See for example htslib_util.h and samfile_util.h. This used to be a much bigger problem before htslib as the
API of samtools was quite incomplete.

I will do this.

jeromekelleher:
Aha, that makes sense. Thanks @AndreasHeger!

PR #94, opened 2015-03-24
get_aligned_pairs_matches_only - faster special case function

TyberiusPrime:
This special case of get_aligned_pairs() offers enormous speed ups when all you care about are SNPs.
One the one hand, you can tell cython to expect integer results from this function, on the other hand you save a ton of time by simply not having (pos, None) for skipped sections (BAM_CREF_SKIP) from your reads in the result list.

In my totally unscientific benchmark here with a coverage per base and position pileup replacement on RNAseq data this adds up to a factor of about 30x.

Best regards
Florian Finkernagel

### DIFF
--- a/pysam/calignmentfile.pyx
+++ b/pysam/calignmentfile.pyx
@@ -3008,6 +3008,42 @@ cdef class AlignedSegment:
 
         return result
 
+    def get_aligned_pairs_matches_only(self):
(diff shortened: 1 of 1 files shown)


TyberiusPrime:
Never mind this pull request - I found some more issues with get_aligned_pairs and will have a more complete patch shortly.

PR #95, opened 2015-03-25
Almost all CIGAR and optional matches_only for read.get_aligned_pairs()

TyberiusPrime:
Get_aligned_pairs was broken for softly clipped reads, and match/mismatch CIGAR commands.
I added test cases for all CIGAR commands.
Padding (6 - BAM_CPAD) is still unsupported, but now raises a NotImplementedError.

Also, I've integrated my recently suggested get_aligned_pairs_matches_only() which allows great speed ups if all you care about is actually read bases.

### DIFF
--- a/pysam/calignmentfile.pyx
+++ b/pysam/calignmentfile.pyx
@@ -2969,8 +2969,14 @@ cdef class AlignedSegment:
 
         return qpos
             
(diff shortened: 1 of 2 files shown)


AndreasHeger:
Thanks!

PR #99, opened 2015-03-31
Query position doc change

TyberiusPrime:
This is the simple doc change that reflects Pileupread.query_positions undefiniedness when there is no query sequence to the alignment at this position.


AndreasHeger:
Thanks!

Issue #93, opened 2015-03-24
Pileup is wrong with spliced reads


TyberiusPrime:
If you have  read spliced across a region, pileup will still happily report reference and query positions
even for positions where the read actually has no aligned base.

Toy example: Should not output anything, outputs [(100, 2), (101, 2)...]

import pysam

sam = pysam.AlignmentFile("shu.bam", 'wb', reference_names = ['chr1'], reference_lengths=[100000])
r = pysam.AlignedSegment()
r.query_name = "test1"
r.seq = "ACCT"
r.qual = "xxxx"
r.pos = 100
r.cigartuples = [ (0,2), (3, 500), (0,2)]
sam.write(r)
sam.close()
pysam.index('shu.bam')

sam = pysam.AlignmentFile('shu.bam')
for p in sam.pileup('chr1', 100, 200, truncate=True, stepper='nofilter'):
    for read in p.pileups:
        print 'reference pos', p.reference_pos, 'query pos', read.query_position

TyberiusPrime:
This is a duplicate of #86

AndreasHeger:
Hi @TyberiusPrime  and @sheitandard, 

when a read is not aligned to a position, the read.is_del flag is set. Will this help you?

I am thinking that there might be use cases where it might be of interest to record reads
spanning a region even if they are not actually aligned.

Best wishes,
Andreas

TyberiusPrime:
It would help you, but it's still a very error prone API, getting the same query position multiple times is unexpected to say the least and leads to subtle bugs when counting coverage.

I guess a behavior like get_aligned_pairs(), returning None for .query_position would be more appropriate.

There are also cases where the query_position is outside of the query string - a 1S50M read for example will lead to a .query_position of '50'.

Though that still does not distinguish between deletions and skipped regions. Samtools pileup doesn't either and counts them both as 'coverage' (an unfortunate decision, IMHO).
What do you think about returning both False (skipped) and None (deletion) for the query position in  get_aligned_pairs()?

Also, I was wondering if you'd be open to have a 'get_coverage(contig, start, stop, consider_read_callback, quality_threshold)' function that returns a set of numpy arrays. My own implementation is about a factor of 56 faster than a straightforward iteration through the pileups (and basically limited by .fetch()).

Obviously it's not as universal as the pileup functionality.

So long,
Florian


Ps: While we're working on this, can we remove the naming difference between reference_pos and query_positions?

sheitandard:
[...]

PR #100, opened 2015-03-31
Pileup read.query position none

TyberiusPrime:
This change returns None for query_position when a PileupRead is a deletion or reference skip at this reference position.
See bug #93


AndreasHeger:
Thanks

PR #101, opened 2015-03-31
Introduced AlignmentFile.count_coverage

TyberiusPrime:
This pull request introduces a function to quickly count coverage in a given region.
I reach about a 26x speed up compared to the pure python implementation.


AndreasHeger:
Great, many thanks!

PR #105, opened 2015-04-11
Add an is_supplementary attribute to AlignedSegment.

marcelm:
This adds the `AlignedSegment.is_supplementary` property. I was not able to run the tests, unfortunately (TestUtils could not be imported and various other errors), but I believe this to be straightforward enough.


AndreasHeger:
Thanks!

marcelm:
Thanks as well :-)

PR #112, opened 2015-04-23
Make AlignedSegments comparable

marcelm:
This adds an implementation of `__richcmp__` so one can compare AlignedSegments to each other. I’m relying on the already existing `compare()` method. Since it’s not clear what the comparison result should be if other operators such as `<` and `<=` are used, I left them out for now.

I still don’t know how to get the unit tests running (as I mentioned in the last pull request), so I haven’t added any.

### DIFF
--- a/pysam/calignmentfile.pyx
+++ b/pysam/calignmentfile.pyx
@@ -2449,6 +2449,14 @@ cdef class AlignedSegment:
             return retval
         return memcmp(t.data, o.data, t.l_data)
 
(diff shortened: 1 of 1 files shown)


marcelm:
Oh, and the reason for adding this was that `==` does not work currently: It returns False even for identical alignments (perhaps it checks for object identity, not equality). I want to fetch() from two nearby positions in a BAM file and be able to tell whether I got the same alignment twice.

marcelm:
Nice!

PR #115, opened 2015-05-19
More work on read/write support, API improvements

bioinformed:
* Add mutable properties to VariantRecord
* Add `phased` property on `VariantRecordSample`
* Release GIL around htslib IO operations


bioinformed:
My next pull request will be to release GIL around the IO operations in the alignment and tabix code.

PR #116, opened 2015-05-20
fetch(reference=...) does not return mates on a different reference

spock:
When using `iter = fetch(reference = "...")` and then iterating over `iter`, paired reads with mates on other references are never encountered even though they exist: `read.reference_id == read.next_reference_id` for **all** reads examined this way.

Iterating on the AlignmentFile itself (using the same file) does find pairs with mates on different references (so there are now reads for which `read.reference_id == read.next_reference_id` is **not** true).

### DIFF
--- a/pysam/calignmentfile.pyx
+++ b/pysam/calignmentfile.pyx
@@ -717,6 +717,10 @@ cdef class AlignmentFile:
 
         Note that a :term:`SAM` file does not allow random access. If
         *region* or *reference* are given, an exception is raised.
+        
+        Also note that using fetch() will not yield reads with mates
+        properly mapped to a different reference: all reads returned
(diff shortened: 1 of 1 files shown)


AndreasHeger:
Many thanks, I was not aware of this!

Best  wishes,
Andreas

spock:
I did some extra testing - and I was wrong, sorry... The true reason appears to be in my use of the `is_proper_pair` read attribute - this is the one which removes any reads with mates on different references; this seems to be the correct/expected behaviour. I had tested this with both `fetch()` and iterating reads from the `AlignmentFile` handle - there is now difference, `fetch()` isn't to blame. I'll now submit a reversion request, sorry for this noise :)

PR #117, opened 2015-05-22
Revert "fetch(reference=...) does not return mates on a different ref…

spock:
Reverts pysam-developers/pysam#116: fetch() works properly, it was is_proper_pair attribute which (by design) filtered out reads with mates on different references.


AndreasHeger:
Thanks!

PR #122, opened 2015-06-10
Fix various spelling errors

afif-elghraoui:
Debian QC tools complained about spelling errors in the source
code. This patch corrects them.


afif-elghraoui:
I'm sorry to bring up something so trivial, but this is something that we have to deal with downstream for the Debian package. If you could please merge this patch, it would save us recurring headaches for future revisions of the package.

AndreasHeger:
On the contrasts, many thanks for your contribution.

PR #127, opened 2015-06-18
Created pFastqProxy and pFastqFile objects for working with multiple fastq records per file

dnephi:
These objects make it possible to work with multiple fastq records per fastq file. With FastqProxy, being a thin wrapper for kseq_t, if you appended records from a fastqfile to a list, you would get a list of FastqProxy objects all pointing to the same kseq_t object.

pFastqProxy has a __str__() function, making it easy to convert back a record into a fastq record string.
Finally, the getQualArray() method seems to be twice as fast as fromQualityString()

Where b is a pFastqProxy object:

In [4]: %timeit fromQualityString(b.quality)
The slowest run took 6.87 times longer than the fastest. This could mean that an intermediate result is being cached 
100000 loops, best of 3: 6.39 µs per loop

In [5]: %timeit b.getQualArray()
The slowest run took 6.88 times longer than the fastest. This could mean that an intermediate result is being cached 
100000 loops, best of 3: 3.64 µs per loop

This is faster because of the new conversion method I added - cs_to_ph(bytes), which treats the characters in byte strings as integers implicitly, removing the need to explicitly convert.

### DIFF
--- a/pysam/cfaidx.pxd
+++ b/pysam/cfaidx.pxd
@@ -54,8 +54,8 @@ cdef class pFastqProxy:
     """
     Python container for pysam.cfaidx.FastqProxy with persistence.
     """
-    cdef public cython.str comment, quality, sequence, name
+    cdef public bytes comment, quality, sequence, name
     cdef cython.str tostring(self)
     cpdef array.array getQualArray(self)
 
-cdef array.array cs_to_ph(cython.str input_str)
+cdef array.array cs_to_ph(bytes input_str)
--- a/pysam/cfaidx.pxd
+++ b/pysam/cfaidx.pxd
@@ -3,9 +3,14 @@ from libc.stdint cimport uint8_t, uint16_t, uint32_t, uint64_t
 from libc.stdlib cimport malloc, calloc, realloc, free
 from libc.string cimport memcpy, memcmp, strncpy, strlen, strdup
 from libc.stdio cimport FILE, printf
+from cpython cimport array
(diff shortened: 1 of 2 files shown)


AndreasHeger:
Thank you very much, this should go in, but I need first some time to review.

AndreasHeger:
Hi, many thanks for this. I propose to add a persist option to FastxFile instead of creating a new container. I have pushed by changes, please let me know your thoughts. I also made persistence to
be the default to make it consistent with AlignedSegment objects, that also persist.

Best,
Andreas

dnephi:
Hi Andreas,

Thanks for taking the time to look this over - this looks good and clean.

The use case I see breaking this is if someone were to use pysam.cfaidx.FastxFile on a fasta file and then iterate through it, the tostring() method would return a 4-line string with "None" (the output of "%s" % None) for quality and have the unnecessary/incorrect "+" line.

I propose adding this to PersistentFastqFile.tostring to prevent that from breaking:

```python
if(self.quality is None):
    return ">%s %s\n%s\n" % (self.name, self.comment, self.sequence)
else:
    return "@%s %s\n%s\n+\n%s\n" % (self.name, self.comment,
                                        self.sequence, self.quality)
```

It's probably an edge case, but worth a thought.

AndreasHeger:
Thanks, you are right of course. I will amend and merge.

Best wishes,
Andreas

AndreasHeger:
Closed, see #134

Issue #130, opened 2015-06-25
failing to print records when parsing vcf via tabix parser


misha-at-genestack:
```python
Traceback (most recent call last):
  File "./pysamvcftest_vcffile.py", line 12, in <module>
    print record
  File "pysam/TabProxies.pyx", line 350, in pysam.TabProxies.TupleProxy.__str__ (pysam/TabProxies.c:4963)
TypeError: decode() argument 1 must be string, not VCF
```

and

```python
Traceback (most recent call last):
  File "./pysamvcftest_vcffile.py", line 12, in <module>
    print record.info
  File "pysam/cvcf.pyx", line 190, in pysam.cvcf.VCFRecord.info.__get__ (pysam/cvcf.c:4933)
  File "pysam/TabProxies.pyx", line 350, in pysam.TabProxies.TupleProxy.__str__ (pysam/TabProxies.c:4963)
TypeError: decode() argument 1 must be string, not VCF
```

AndreasHeger:
Add unit test to cover this.

AndreasHeger:
Done, see d26cb3818a9488bf5c4755deb07d94a91a35f82e

PR #131, opened 2015-06-25
Fix for #130

misha-at-genestack:
I believe this fixes issue #130, passing encoding correctly to `TabProxies.TupleProxy`.


AndreasHeger:
Thanks!

PR #132, opened 2015-06-25
get_tag: expose tag type if requested with `with_value_type`

dalexander:
Would you be receptive to a change along these lines?  Getting the type of a tag value is important to us.  I can add some tests.


AndreasHeger:
Thanks!

Issue #125, opened 2015-06-15
Serialize AlignedSegment to SAM string


danvk:
[According to the code][1], the `str` representation of AlignedSegment "is an approximate :term:`sam` format."

I believe the only way to get an exact SAM format is to serialize the alignment to a temp file on disk and then read it back. Is this correct? If so, it would be nice to have a more direct way.

(context: I'd like to read segments of a BAM file using pysam and send them over the network in a standard format.)

[1]: https://github.com/pysam-developers/pysam/blob/d3b8556dbafe62142f5a2bca78565ae9e58fae24/pysam/calignmentfile.pyx#L2465

AndreasHeger:
Hi @danvk , the problem with AlignedSegment is that these objects can live in a context that is independent of a SAM-file. The SAM-file keeps track of chromosome names, while the AlignedSegment contains only an index of the chromosomal identifier (tid). Hence, the only time an AlignedSegment
can be written as a full SAM entry is within the context of a SAM-file (AlignmentFile). Thus, str can never be complete SAM format.
It might be worth adding a json dump to AlignedSegment.

AndreasHeger:
Thanks, tostring() should cover this use case.

PR #140, opened 2015-07-03
AlignedSegment to valid SAM record string

dnephi:
Added tostring() method to AlignedSegment using its matched AlignmentFile. (See https://github.com/pysam-developers/pysam/issues/125)

This is motivated by issues with AlignmentFile's flexibility in writing - it fails to write the headers for its plain text files. You can manually write it to a stream (handle.text), but not to a file because AlignmentFile doesn't have a "w+" option, so any file header gets clobbered (and so the answer is a stream).

I also often get memory allocation issues when moving AlignedSegment_t objects around, and so when I've processed reads but sometimes want to output my new text or the read's texts, I end up with two streams. (In fact, the AlignedSegment from scratch example in the docs works for me to write, but segfaults when I try to read that file.) Of course, piped output is important for performance, and the way that pipes work is that they close the moment they receive a close message, making it impossible to force both textual output through a stdout stream and AlignmentFile("-") into the same process. (You could make an ugly hack, write one to stderr, and merge them, but that's disgusting and dangerous.)

The perfect answer for me is having a valid tostring method so that I can write them both to the same file/stream. I figure it'd be very useful to be able to append sam strings to files, giving the records a bit more mobility.

Also, added the method QualStringFromArray, which takes advantage of python's array('B').tostring() method and python's incredible fast string translate() function. Pre-computing the string.maketrans matrices makes this the fastest way I've been able to offset quality scores in a python array.

If toQualString receives a python array as an argument, it will execute the faster, more strongly-typed code.

### DIFF
--- a/pysam/calignmentfile.pxd
+++ b/pysam/calignmentfile.pxd
@@ -7,6 +7,12 @@ from libc.stdio cimport FILE, printf
 from cfaidx cimport faidx_t, Fastafile
 from chtslib cimport *
 
+cimport cpython.array
+cimport cython
+
+ctypedef cpython.array.array py_array_t
+ctypedef AlignmentFile AlignmentFile_t
+
 cdef extern from *:
     ctypedef char* const_char_ptr "const char*"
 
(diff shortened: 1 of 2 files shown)


dnephi:
I did also make a few changes to improve PEP8 compliance. If it is preferable, I can roll back the whitespace modifications I made.

AndreasHeger:
Dear @dnephi, many thanks. I have added tostring, which will be very useful. I changed the implementation for the binary tag conversion. Yours was very clean and elegant, but using memcpy instead of elementwise copying turned out to be significantly faster. I hope you don't mind.

Best wishes,
Andreas

PR #142, opened 2015-07-09
Fix the unmapped count, incl reads with no SQ group

nlhepler:


nlhepler:
Presently, f.unmapped does not count unmapped records with no SQ group (aka tid = -1).

AndreasHeger:
Many thanks

PR #147, opened 2015-07-14
Patch for an edge case in array packaging/unpackaging.

dnephi:
Apparently, an array of length 1 without a terminating comma is viewed as an incompleted aux field. Modified TagToString for this edge case.

### DIFF
--- a/pysam/calignedsegment.pyx
+++ b/pysam/calignedsegment.pyx
@@ -170,35 +170,20 @@ cdef bytes TagToString(tuple tagtup):
         min_value = min(b_aux_arr)
         length = len(b_aux_arr)
         if(size == 1):
-            if(length != 1):
-                if(min_value < 0):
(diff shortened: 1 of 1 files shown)


dnephi:
I checked the Travis build, and it looks like it failed for IOError reasons. Any idea on that?

=====================================================================

ERROR: testAsString (AlignedSegment_test.TestAsString)

----------------------------------------------------------------------

  File "/home/travis/build/pysam-developers/pysam/tests/AlignedSegment_test.py", line 477, in testAsString

    with open(os.path.join(DATADIR, "ex2.sam")) as samf:

IOError: [Errno 2] No such file or directory: 'pysam_data/ex2.sam'

AndreasHeger:
There was an issue with sourceforge (needed to install samtools), I think it has been fixed.
I have restarted the tests.
Best wishes,
Andreas

AndreasHeger:
Thanks

Issue #71, opened 2015-01-15
ValueError: unknown field code 'PM' in record 'RG'


cc2qe:
Thanks for maintaining this great piece of software Andreas.

I get an error with this example script and SAM file below. It appears that my BAM uses an unrecognized tag in the header. I realize that "PM" is not in the SAM spec, but is it possible to make pysam more lenient instead of crashing? In general I do not expect pysam to validate my BAM files, and this is also a TCGA file downloaded from CGHub.

```
$ ./parse_rg.py
Traceback (most recent call last):
  File "./parse_rg.py", line 7, in <module>
    for tag in bam.header['RG']:
  File "csamfile.pyx", line 1052, in pysam.csamfile.Samfile.header.__get__ (pysam/csamfile.c:12899)
ValueError: unknown field code 'PM' in record 'RG'
```

parse_rg.py:
```python
#!/usr/bin/env python

import pysam

bam = pysam.Samfile("example.sam", 'r')

for tag in bam.header['RG']:
    print tag
bam.close()
```

example.sam:
```
@HD   VN:1.5  SO:coordinate
@SQ   SN:1    LN:249250621
@SQ   SN:2    LN:243199373
@SQ   SN:3    LN:198022430
@RG   ID:BCM:0    PL:ILLUMINA CN:BCM  DT:2012-08-24T21:00:00Z LB:WGS:BCM:IWG_THNSCG.CV-7090-10AN_1pA  PM:Illumina HiSeq 2000  SM:mysample PU:BCM:120807_SN601_0319_BC10LPACXX_8
@RG   ID:BCM:0_1  PL:ILLUMINA CN:BCM  DT:2012-12-08T10:00:00Z LB:WGS:BCM:IWG_THNSCG.CV-7090-10AN_1pA  PM:Illumina HiSeq 2000  SM:mysample PU:BCM:121117_SN898_0330_BD1DYHACXX_7_ID05
@RG   ID:BCM:0_1_1    PL:ILLUMINA CN:BCM  DT:2012-09-04T19:00:00Z LB:WGS:BCM:IWG_THNSCG.CV-7090-10AN_1pA  PM:Illumina HiSeq 2000  SM:mysample PU:BCM:120822_SN580_0246_AC119AACXX_8
[...]

PR #153, opened 2015-08-05
Removes exception for custom upper-case header record tags.

MurphyMarkW:
The SAM specification advises as best practice to avoid use of custom upper-case header record tags without requesting their addition to the specification, but the spec does not preclude their addition or use. This pull request simply alters the header record parsing behavior to support this.

Happens to be related to #71.

o7


AndreasHeger:
Many thanks

PR #159, opened 2015-08-18
Added function signatures to cvcf cython.  Added note about other VCF code.

kyleabeauchamp:
The rest of the changes are whitespace trimming, sorry about that.  The only real changes  should be at the top of the file.  Before and after docstrings are shown below.  Not sure how you feel about my additional note in the comments, but I had no idea about the difference between VariantFile and VCF until I looked through the source code---I thought a note would be in order.

```
v = pysam.VCF()
v.write?
Docstring: Writes a VCF file to a stream, using a data generator (or list)
Type:      instancemethod

[...]

v = pysam.VCF()
v.write?
Docstring:
VCF.write(self, stream, datagenerator)
Writes a VCF file to a stream, using a data generator (or list)
Type:      instancemethod
```

### DIFF
--- a/pysam/cvcf.pyx
+++ b/pysam/cvcf.pyx
@@ -1,3 +1,4 @@
+# cython: embedsignature=True
 #
 # Code to read, write and edit VCF files
 #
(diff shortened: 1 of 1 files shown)


AndreasHeger:
Many thanks, this is a neglected part of the code. 
@bioinformed is working on bcf/vcf support through htslib.

PR #166, opened 2015-09-11
adds context manager functions to FastaFile

jeffhussmann:
Allows FastaFile to be used in a 'with' statement.


AndreasHeger:
Many thanks

PR #169, opened 2015-09-23
Add reference_name and next_reference_name attributes to AlignedSegment

marcelm:
PileupColumn also gets a reference_name attribute. For this to work,
both AlignedSegment and PileupColumn gain an _alignment_file attribute.

No need to call getrname anymore in user code!

I cannot get the tests to run locally, so there aren’t any test cases, yet. I’m opening this PR already now to get feedback from Travis.

Another possibility of implementing this would be to only have a reference_name attribute, but the approach with the alignment_file attribute is more generic. For example, I haven’t implemented a setter for reference_name, yet, but to do that, it will also be necessary to access the AlignmentFile.

### DIFF
--- a/pysam/calignedsegment.pxd
+++ b/pysam/calignedsegment.pxd
@@ -43,6 +43,9 @@ cdef class AlignedSegment:
     # object that this AlignedSegment represents
     cdef bam1_t * _delegate
 
+    # the file from which this AlignedSegment originates (can be None)
+    cdef AlignmentFile _alignment_file
+
     # caching of array properties for quick access
     cdef object cache_query_qualities
     cdef object cache_query_alignment_qualities
@@ -72,6 +75,7 @@ cdef class PileupColumn:
     cdef int tid
     cdef int pos
     cdef int n_pu
+    cdef AlignmentFile _alignment_file
 
 
 cdef class PileupRead:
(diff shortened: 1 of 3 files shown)


AndreasHeger:
Many thanks. There have been discussions in the past about a reference_name and my gut feeling has always been to not link AlignmentFile and AlignedSegment implicitely as the latter can exist without
the former, or an AlignedSegment might move between two files. 

I think it best to link explicitely. Would a function get_reference_name(AlignmentFile) work for you, similar to AlignedSegment.tostring(AlignmentFile)?

Best wishes,
Andreas

marcelm:
If an AlignedSegment is not linked to an AlignmentFile, then the reference_name property will return None, so this is taken care of.

Also, if you move an AlignedSegment to a different AlignmentFile, then you already have the problem that the reference_id refers to an id in a different file, so there wouldn’t be more problems than you already have now. I’d be happy to add a sentence about that to the documentation.

I would also assume that pysam users mostly iterate over files, run pileups and perhaps modify the reads a bit. In those cases, it would just work. My motivation is that I often write functions that work on single reads. All these functions require that I pass in both the AlignedSegment and the AlignmentFile if I need the reference name anywhere - and even if it’s just for debugging purposes.

get_reference_name(AlignmentFile) isn’t better than the current solution.

I believe the reference_id thing is a BAM or htslib implementation detail that is leaking through, while it could easily be hidden from users, especially in a Python library where usability counts.

AndreasHeger:
Ok. thanks, will give it ok.

marcelm:
Cool, thanks for reconsidering! And sorry for nagging ... Note there aren’t any tests, yet. I hope I can send you a separate PR for them soon.

PR #180, opened 2015-11-09
Fixes AlignmentFile.check_index to return True

dannon:
(when not raising an exception) as defined in the docstring.


AndreasHeger:
Thanks!

PR #188, opened 2015-12-01
Fix docstring in aligned pairs

kyleabeauchamp:
Currently the docs on `get_aligned_pairs()` appear misformatted (http://pysam.readthedocs.org/en/latest/api.html#api).  I think my changes fix them.  The remaining change should just be whitespace cleanup by my editor.


kyleabeauchamp:
Also, does the description for `matches_only` seem counterintuitive?  It seems to suggest that it will only return REF matches, but I that's not quite true...

AndreasHeger:
Thanks!

PR #193, opened 2015-12-16
Do not release GIL when querying Tabix

hannes-brt:
Tabix is not thread-safe and the GIL should not be released while querying.

In my case, I was running a Tornado web server with several threaded workers that performed Tabix queries simultaneously.

This would result in error messages from the Tabix C-library such as this:
```
[E::get_intv] failed to parse TBX_GENERIC, was wrong -p [type] used?
The offending line was: "	+	NM_012231	6	intronic	CDS	238	682833521792	34873960360784	iffffffffffffffffff”
```

The error message indicated that lines from the indexed file where read incompletely and the issue disappeared when I switched to using just a single worker.

The attached patch solved the issue for me.

### DIFF
--- a/pysam/ctabix.pyx
+++ b/pysam/ctabix.pyx
@@ -421,16 +421,14 @@ cdef class TabixFile:
 
         if region is None:
             # without region or reference - iterate from start
-            with nogil:
-                iter = tbx_itr_queryi(fileobj.index,
-                                      HTS_IDX_START,
-                                      0,
-                                      0)
+            iter = tbx_itr_queryi(fileobj.index,
+                                  HTS_IDX_START,
+                                  0,
+                                  0)
         else:
             s = force_bytes(region, encoding=fileobj.encoding)
             cstr = s
-            with nogil:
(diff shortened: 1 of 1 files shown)


mp15:
Might be worth logging a feature request with htslib for this. I assuming you're using the hts_index to access Tabix rather than invoking the tool?

hannes-brt:
I am using `pysam.TabixFile`.

AndreasHeger:
Many thanks, I will watch this.

bioinformed:
Pretty sure that not releasing the GIL is the wrong fix.  I don't *think* the tabix code uses global state, so the thread unsafety is for shared access to a single index.  A better approach is to document the thread-unsafety and services should implement their own granular locks, rather than serializing all Tabix IO under the GIL.

bioinformed:
Also, iter is not a reserved word in Python.  It is a built-in function that probably shouldn't be aliased by a local variable.

AndreasHeger:
Reverting this change and adding a section to the faq.

I have raised a new issue, there might be different, thread-safe path to iterate through the htslib API.

PR #199, opened 2016-01-08
Incorrect cram file name trim on index discovery

jwhsanger:
Just eyeballing, this looks like a minor bug - please reject if irrelevant!


AndreasHeger:
Thanks

Issue #197, opened 2016-01-05
TagToString fails with "TypeError: expected bytes, str found"


humburg:
I'm trying to use the *tostring* method of *AlignedSegment* to get a string representation of the read alignment (using Python 3.4.3). If the alignment object contains any optional tags this fails with

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "pysam/calignedsegment.pyx", line 739, in pysam.calignedsegment.AlignedSegment.tostring (pysam/calignedsegment.c:10523)
  File "pysam/calignedsegment.pyx", line 773, in pysam.calignedsegment.AlignedSegment.tostring (pysam/calignedsegment.c:10383)
  File "pysam/calignedsegment.pyx", line 779, in pysam.calignedsegment.AlignedSegment.get_tag_string (pysam/calignedsegment.c:10644)
  File "pysam/calignedsegment.pyx", line 139, in pysam.calignedsegment.TagToString (pysam/calignedsegment.c:3674)
TypeError: expected bytes, str found
```

Alignments without tags work fine.

humburg:
On closer inspection alignments without tags are broken as well. They don't cause an exception when calling *tostring* but they have corrupted base qualities. Looks like there is some confusion over the use of `str` and `bytes` in *tostring*.

AndreasHeger:
Thanks, I will have a look!

NoSeatbelts:
I wrote TagToString and tostring, and I wrote them primarily concerned with python 2.

I've modified the methods to not use cython.str declarations.

humburg, would you see if the issue's been fixed for you?

https://github.com/pysam-developers/pysam/pull/200

humburg:
@NoSeatbelts, Thanks for looking at this. I tried to test your fix but I get

```
86_64-linux-gnu-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -g -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2 -fPIC -D_FILE_OFFSET_BITS=64 -D_USE_KNETFILE= -Isamtools -Ipysam -Ihtslib -I/usr/include/python3.4m -c pysam/csamtools.c -o build/temp.linux-x86_64-3.4/pysam/csamtools.o -Wno-error=declaration-after-statement -DSAMTOOLS=1
[...]

PR #200, opened 2016-01-12
Bytes

NoSeatbelts:
In response to [this issue](https://github.com/pysam-developers/pysam/issues/197), cython.str declarations removed and replaced with <bytes> coercions.


AndreasHeger:
Many thanks, but using a different solution through using sam_write1 API function.

PR #212, opened 2016-01-22
Fix rendering error on readthedocs

gshiba:
The two for loops were rendered as separate loops instead of being nested (at http://pysam.readthedocs.org/en/latest/api.html)


AndreasHeger:
Thanks

PR #214, opened 2016-01-29
get_tags() should return the tag type as a character, not an integer

dpryan79:
Currently `get_tags(with_value_type=True)` returns a tuple of (two characters, value, integer type). So an auxiliary tag like `AS:i:1` is returned as `('AS', 1, 67)` rather than `('AS', 1, 'C')`. This PR just puts things through `chr()`, which should solve the problem. I assume that this issue only arises due to how Cython dealt with the code previously.

### DIFF
--- a/pysam/calignedsegment.pyx
+++ b/pysam/calignedsegment.pyx
@@ -1852,7 +1852,7 @@ cdef class AlignedSegment:
             s += 1
 
             if with_value_type:
-                result.append((charptr_to_str(auxtag), value, auxtype))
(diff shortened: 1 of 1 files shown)


dpryan79:
I can confirm that Cython returns a uint8_t for this otherwise, which makes sense given how the `bam1_t` struct works.

JohnLonginotto:
I just came here to complain about the same thing, but it looks like you beat me too it dpryan79! ;)
Thank you!! :) Looking forward to this being implemented in stable!

PR #216, opened 2016-02-01
Add write support for FILTER, INFO and FORMAT, with some other minor fixes

bioinformed:
In addition to adding write support for INFO and FORMAT, all non-scalar INFO and FORMAT values (Number!=1) are now returned as tuples.


bioinformed:
Added support for deleting format values from a sample, entirely removing a format from a record and removing info from a record.

Removed support for indexing INFO and FORMAT as a sequence (by integer index), since deletion leaves gaps.  Fixed lookups, iteration and length determination for INFO and FORMAT to ignore gaps.

Additional fixes include making VariantFile.close() work again for non-compressed output.  The previous fix only worked for compressed output and is no longer needed, regardless, due to better error reporting in VariantFile.open() added in my last PR.

Temporarily removed the check on writing records derived from a different header.  Without implementing the translation system, this effectively disables all writing.  I will re-enable this check once I figure out how to safely use the htslib translation mechanism or implement a more robust version.  The surface issue with the htslib version is that a single translation cache is stored in the source header and there is no apparent check whether the source header is used with multiple destination headers.  Thus any application that splits a single variant file into multiple with different destination headers can result in incorrect output.

It bit the bullet and finally installed a modern Py3 VE and have been fixing emergent str/bytes issues.

bioinformed:
Current breakage in tabix_test.py is semi-intentional and will be addressed in my next push.

AndreasHeger:
Thanks, that is really great! Need a few evenings off, but happy to do some testing, let me know when a good point has been reached.

AndreasHeger:
Great, thanks!

bioinformed:
Thanks for merging.  I have a couple more minor features and a bunch of test code yet to add.  These should make cbcf feature complete, at least to the point where we can start to get serious feedback on the API choices.  Can you wait just a few more days before making the 0.8.5 release to give me time to finish?

AndreasHeger:
Certainly!

I have added a couple of tests for creating VCF files from scratch, some of which cause failures. There are also a couple of tabix tests still failing - I will see if I can fix them and submit them as a pull request. Could you please review them?

PR #220, opened 2016-02-05
Update cy_build.py to support pip and develop mode on Mac.

mdpearson:
Hello,

I'm a software developer at Invitae, and we've been using pysam for quite some time. Recently I saw an update on the 0.8.5-dev branch that supports linking to pysam .so files on Mac. This made it possible for us to port an internal library to Mac. In the course of that project, I modified cy_build.py to:

a. work with pip (the current implementation depends on easy_install's egg placement)
b. work with develop mode (either `python setup.py develop` or `pip install -e`)
c. use the `@rpath` flag within the clang linker's `install_name` so shared libraries do not contain the full path of the virtual environment, if any, in which they were built.

I hope these changes are helpful, please reach out to me if you have any questions, suggestions for improvement, etc. Keep up the great work.

### DIFF
--- a/pysam/__init__.py
+++ b/pysam/__init__.py
@@ -43,9 +43,26 @@ from pysam.version import __version__, __samtools_version__
 def get_include():
     '''return a list of include directories.'''
     dirname = os.path.abspath(os.path.join(os.path.dirname(__file__)))
-    return [dirname,
-            os.path.join(dirname, 'include', 'htslib'),
-            os.path.join(dirname, 'include', 'samtools')]
+
+    #
+    # Header files may be stored in different relative locations
+    # depending on installation mode (e.g., `python setup.py install`,
+    # `python setup.py develop`. The first entry in each list is
+    # where develop-mode headers can be found.
+    #
+    htslib_possibilities = [os.path.join(dirname, '..', 'htslib'),
+                            os.path.join(dirname, 'include', 'htslib')]
+    samtool_possibilities = [os.path.join(dirname, '..', 'samtools'),
+                             os.path.join(dirname, 'include', 'samtools')]
+
+    includes = [dirname]
+    for header_locations in [htslib_possibilities, samtool_possibilities]:
+        for header_location in header_locations:
+            if os.path.exists(header_location):
+                includes.append(os.path.abspath(header_location))
+                break
+
+    return includes
 
 
 def get_defines():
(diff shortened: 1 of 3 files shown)


pysam/__init__.py
```diff
@@ -43,9 +43,26 @@
...
-            os.path.join(dirname, 'include', 'samtools')]
+
+    #
+    # Header files may be stored in different relative locations
+    # depending on installation mode (e.g., `python setup.py install`,
+    # `python setup.py develop`. The first entry in each list is
+    # where develop-mode headers can be found.
+    #
+    htslib_possibilities = [os.path.join(dirname, '..', 'htslib'),
+                            os.path.join(dirname, 'include', 'htslib')]
```
jmarshall: For htslib, code should be using `#include <htslib/foo.h>`, so the correct directory to add to an include path is the parent of a _htslib_ directory, not that directory itself.

mdpearson: Turns out, in my first attempt to support develop-mode I did exactly as you suggested. Somewhat confusingly, though, _htslib_ header files are stored in a subdirectory called `htslib`, twice... For example, when you run `python setup.py develop`, htslib headers can be found in `<$PKG_ROOT/htslib/htslib/hts.h>`, while when you run `python setup.py install` they live in, e.g., `$YOUR_VE/lib/python2.7/site-packages/pysam/include/htslib/htslib/hts.h`.

AndreasHeger:
Thanks, @mdpearson, that is great! I will try to get travis OSX to run and then merge.

AndreasHeger:
Ok, OSX and python on travis is not going to happen..., see https://docs.travis-ci.com/user/languages/python

kyleabeauchamp:
AFAIK, OSX + Travis could be achieved by using the `miniconda` approach that is used by bioconda:

https://github.com/bioconda/bioconda-recipes/blob/master/scripts/travis-setup.sh

AndreasHeger:
Thanks, I will give it a go - do I need to change the language from python to C? This seems to be necessary in all the workarounds I have seen.

kyleabeauchamp:
I think so:

https://github.com/bioconda/bioconda-recipes/blob/master/.travis.yml

bioinformed:
@AndreasHeger @mdpearson This PR looks extremely promising.  Since I do most of my development on OSX, I'm happy to test as well.

pysam/__init__.py
jmarshall: Fair enough, so the parent of the _htslib_ directory happens to be called _htslib_ itself.  Not sure that that makes for the ideal `include` directory under _site-packages/pysam_, but this is not my domain…

PR #224, opened 2016-02-09
Add missing cbcf functionality

bioinformed:
This PR adds the most glaring missing functionality from cbcf -- getting and setting genotypes on samples.  It also adds:
- id/key string caching/interning
- support for Flag types
- setting alleles from any mixture of indices and strings
- a few more Py3 fixes

tests/tabix_test.py no longer fails, provided remote data are available.

@AndreasHeger : Please merge prior to release of 0.8.5.  

My next goals are to:
- translate my large suite of ad hoc testing code into UnitTests (ETA later this week)
- support header translation to allow records from one file (with one header) to be written to another file (with a different header) (ETA next week)
- complete mutating support by allowing creation of new records and adding samples to existing records (ETA next week)
- add support to htslib and pysam for UTF-8 encoded files (ETA in the coming month or so)
- add support to htslib and pysam for percent encoded INFO and FORMAT values.  This will safely allow commas, semi-colons, colons, tabs, and other currently illegal characters in string values. (ETA in the coming month or so)

### DIFF
--- a/pysam/cbcf.pyx
+++ b/pysam/cbcf.pyx
@@ -5,13 +5,14 @@
 ## Cython wrapper for htslib VCF/BCF reader/writer
 ###############################################################################
 #
-# NOTICE: This code is incomplete and preliminary.  It does offer a nearly
-#         complete immutable Pythonic interface to VCF/BCF metadata and data
-#         with reading and writing capability, but has no capability (yet)
-#         to mutate the resulting data (beyond dropping all samples).
-#         Documentation still needs to be written and a unit test suite is
-#         in the works.  The code is also superficially specific to Python 2
-#         and will require a bit of work to properly adapt to Python 3.
+# NOTICE: This code is incomplete and preliminary.  It offers a nearly
+#         complete Pythonic interface to VCF/BCF metadata and data with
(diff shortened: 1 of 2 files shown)


AndreasHeger:
Great, thanks! This is really amazing work. When do you think is a good time to release - after the next set of changes you have outlined above?

We need to make sure that the API gets proper exposure in the documentation so that people will pick it up. I am happy to start with some documentation. Do you think it is ready, or should we wait until the next release and leave it for now to be discovered by advanced users?

bioinformed:
If you have time, please start on the documentation!  The object structure
and most of the conventions feel fairly mature and stable.  Getting them
written down is necessary and I'd very much appreciate the extra scrutiny
on the aesthetic choices that I made.  My feeling is that we should release
when the documentation is ready and whatever features that lag will catch
up in the next release.

I'd still make it clear that the API and implementation are still in Beta
and are subject to change, but at least we'll get wider exposure to users
and their feedback.

Thanks,
-Kevin

AndreasHeger:
I have made the switch to the cbcf API, mostly for reading, and I find it intuitive.
I will start on documentation in the next few days.

PR #248, opened 2016-03-07
Changes necessary for conda build

kyleabeauchamp:
1.  Added (trivial) support for `--old-and-unmanageable` build option that avoids egg-style naming.
2.  Added workarounds in build.sh for manually re-naming rpaths on OSX conda builds
3.  Added extra RPATH options in meta.yaml, which fixes the conda builds on linux systems

Once we're happy with this, I'll make a PR to bioconda-recipes that points to the latest git commit.  I think it's fine to just call that "0.9.0" for the meantime.

### DIFF
--- a/ci/conda-recipe/build.sh
+++ b/ci/conda-recipe/build.sh
@@ -2,7 +2,36 @@
 
 # Use internal htslib
 chmod a+x ./htslib/configure
-export CFLAGS="-I${PREFIX}/include/curl/ -I${PREFIX}/include -L${PREFIX}/lib"
+export CFLAGS="-I${PREFIX}/include -L${PREFIX}/lib"
 export HTSLIB_CONFIGURE_OPTIONS="--disable-libcurl"
 
-$PYTHON setup.py install
+# Use --old-and-unmanageable to disable egg-style folder naming, which is a minor hassle for RPATH stuff.
+$PYTHON setup.py install --old-and-unmanageable
+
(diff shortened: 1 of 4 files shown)


kyleabeauchamp:
I imagine this could be done more cleanly by refactoring some of the buildchain in setup.py and cy_build.py, but the changes here are "minimally perturbative" of the non-conda build path, which is probably a good thing for now.

kyleabeauchamp:
This would also imply that the conda build that bioconda rolls out would have libcurl disabled.  However, I think that's fine for the time being, right?

sebastian-luna-valero:
Hi Kyle,

I found this thread while I was creating a pysam recipe myself.

A possible workaround is to update setup.py in build.sh as follows:
```sed -i'' -e 's/HTSLIB_MODE = \"shared\"/HTSLIB_MODE = \"separate\"/g' setup.py```

I think the cause of the problem is having HTSLIB_MODE enforced to "shared" in setup.py and then not being able to link with a shared htslib installation:
https://github.com/pysam-developers/pysam/blob/v0.9.0/setup.py#L79

Meanwhile, I have a conda recipe that seems to work:
https://github.com/bioconda/bioconda-recipes/pull/969

Best regards,
Sebastian

kyleabeauchamp:
Looks like your recipe is working, so perhaps I don't need the current changes merged.  Not sure.

PR #258, opened 2016-03-23
Fix test on py3k

kyleabeauchamp:
It seems my test wasn't quite py3k compatible.  Let's see if this does the trick.

### DIFF
--- a/tests/test_samtools_python.py
+++ b/tests/test_samtools_python.py
@@ -16,7 +16,7 @@ def test_bedcov_split_lines():
     for line in lines:
         print(line)
         print(type(line))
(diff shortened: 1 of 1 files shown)


kyleabeauchamp:
So the issue here is that the samtools calls return bytes on py3k and strings on py2k.  This means that users (and tests) cannot do `line.split("\")` because the tab is a string, not a byte object.  Not sure what the right answer is longterm, but at least this fixes the tests.

kyleabeauchamp:
The test failure appears to be a random unrelated failure:

```
ERROR: testRemoteExplicitIndexBAM (AlignmentFile_test.TestExplicitIndex)
```

AndreasHeger:
Thanks!

Issue #262, opened 2016-03-30
pip install pysam error


thkitapci:
Hi, 
I am trying to install pysam with pip I have Python 2.7.11 :: Anaconda 2.5.0 (64-bit)

when I run pip install pysam I get the following output

Collecting pysam
  Using cached pysam-0.9.0.tar.gz
Building wheels for collected packages: pysam
  Running setup.py bdist_wheel for pysam: started
[?25l  Running setup.py bdist_wheel for pysam: finished with status 'error'
  Complete output from command /home/cmb-panasas2/tkitapci/libraries/lib_for_python/anaconda2_2_5_0/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-Lg2cSy/pysam/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" bdist_wheel -d /tmp/tmpIL7lClpip-wheel- --python-tag cp27:
  # pysam: htslib mode is shared
  checking for gcc... gcc
  checking whether the C compiler works... yes
  checking for C compiler default output file name... a.out
  checking for suffix of executables...
  checking whether we are cross compiling... no
  checking for suffix of object files... o
  checking whether we are using the GNU C compiler... yes
  checking whether gcc accepts -g... yes
  checking for gcc option to accept ISO C89... none needed
  checking for ranlib... ranlib
  checking how to run the C preprocessor... gcc -E
  checking for grep that handles long lines and -e... /bin/grep
  checking for egrep... /bin/grep -E
  checking for ANSI C header files... yes
  checking for sys/types.h... yes
  checking for sys/stat.h... yes
  checking for stdlib.h... yes
  checking for string.h... yes
  checking for memory.h... yes
  checking for strings.h... yes
  checking for inttypes.h... yes
  checking for stdint.h... yes
  checking for unistd.h... yes
  checking for stdlib.h... (cached) yes
  checking for unistd.h... (cached) yes
  checking for sys/param.h... yes
  checking for getpagesize... yes
  checking for working mmap... yes
  checking for gmtime_r... yes
  checking for zlib.h... yes
  checking for inflate in -lz... yes
  checking for curl_easy_pause in -lcurl... yes
  checking for CCHmac... no
  checking for library containing HMAC... -lcrypto
  configure: creating ./config.status
  config.status: creating config.mk
  config.status: creating config.h
  config.status: config.h is unchanged
[...]

PR #264, opened 2016-04-01
#262 Fix install issue

Bratdaking:
Tabixproxies and vcf couldn't locate the htslib dirs during a manual install. Added these to the respective parts in the setup script.


AndreasHeger:
Thanks

PR #276, opened 2016-05-06
Update INSTALL

kyleabeauchamp:
Found it


AndreasHeger:
Thanks, missed that one - should have grepped through all files...

PR #282, opened 2016-05-11
Allow specifying a FASTA reference when reading a CRAM file

### LINKED ISSUE #280: Specify path to reference when opening CRAM
`AlignmentFile` has an argument `reference_filename` that allows to specify the name of a FASTA file to be used as the reference when opening a CRAM file for *writing*. I’d like for this option to also work when *reading* CRAM files. Something like the `-T` option for `samtools view`.

I’ve tried to copy and paste the appropriate lines in `Alignmentfile._open` in `calignmentfile.pyx` from the writing branch of the big `if` statement to the reading branch, but it did not work. I’d be happy to submit a PR myself if anyone could point me into the right direction.

marcelm:
The reference_filename parameter to the AlignmentFile constructor is now
also interpreted when opening a CRAM file for reading, not only when writing.

Closes #280

### DIFF
--- a/pysam/calignmentfile.pyx
+++ b/pysam/calignmentfile.pyx
@@ -219,7 +219,7 @@ cdef class AlignmentFile:
     """AlignmentFile(filepath_or_object, mode=None, template=None,
     reference_names=None, reference_lengths=None, text=NULL,
     header=None, add_sq_text=False, check_header=True, check_sq=True,
-    filename=None)
+    reference_filename=None, filename=None)
 
     A :term:`SAM`/:term:`BAM` formatted file. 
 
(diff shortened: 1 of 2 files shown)


marcelm:
There is one problem that should possibly be documented/fixed before this is merged: If `AligenmentFile.fetch()` is called with `multiple_iterators=True`, then the CRAM file is re-opened, but without setting the reference.

AndreasHeger:
Many thanks. I have added reference_filename to VariantFile. For release 0.9.2 I plan to refactor VariantFile to make it a factory function returning different objects for BAM and CRAM files to reflect the additional parameterization of CRAM.

marcelm:
Cool, thanks! You mean AlignmentFile instead of VariantFile, right?

PR #284, opened 2016-05-11
Add a TestDoubleFetchCram unit test

marcelm:
This pull request commits a failing unit test that should in my opinion pass. (It is not meant for merging directly.)

The test is a copy of the TestDoubleFetch test, but it works on `ex1.cram` instead of `ex1.bam`. It tests whether fetch()ing from a CRAM file with multiple iterators works. 

I’m not sure what the problem is, but it fails in the `testDoubleFetchWithRegion` method on the first iteration of the `for` loop. The first `.fetch()` (stored in `a`) correctly returns the first read of the region, but the second `.fetch()` (stored in `b`, with `multiple_iterators` set to True) returns the first read *of the file*.

### DIFF
--- a/tests/AlignmentFile_test.py
+++ b/tests/AlignmentFile_test.py
@@ -1840,7 +1840,7 @@ class TestBTagBam(TestBTagSam):
     filename = os.path.join(DATADIR, 'example_btag.bam')
 
 
-class TestDoubleFetch(unittest.TestCase):
+class TestDoubleFetchBam(unittest.TestCase):
 
     '''check if two iterators on the same bamfile are independent.'''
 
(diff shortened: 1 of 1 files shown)


marcelm:
Here is a simpler test:
```
>>> import pysam
>>> f = pysam.AlignmentFile('pysam_data/ex1.cram')
>>> next(f.fetch('chr1', 200)).qname
'EAS1_103:4:294:525:849'
>>> next(f.fetch('chr1', 200, multiple_iterators=True)).qname
'EAS56_57:6:190:289:82'
```
`...:849` is correct. `...:82` is the first read in the file.

AndreasHeger:
Thawks, will take a look at this this w/e.

Best wishes,
Andreas

AndreasHeger:
Thanks, I have opened an issue and incorporated the unit test.

PR #297, opened 2016-06-15
Update INSTALL

kyleabeauchamp:
AFAIK some info here was previously wrong.


kyleabeauchamp:
PS: unrelated, but looks like the bioconda builds for 0.9.1 are available online:

https://anaconda.org/bioconda/pysam/files

PR #300, opened 2016-06-19
Add bioconda instructions?

kyleabeauchamp:
Given that people are currently having lots of install issues, it may make sense to direct people to the bioconda packages.  Up to you.


AndreasHeger:
Thanks!

PR #301, opened 2016-06-20
Improve handling of build system for external htslib

afif-elghraoui:
This is a patch I have applied in order to build the Debian package using the system htslib installation. I've tried to make it flexible so that it still works for the other two cases you support.


AndreasHeger:
Thanks!

PR #309, opened 2016-07-01
Update calignmentfile.pyx

odoublewen:
Error message ("file has no sequences defined") was misleading.  The argument is check_sq, not check_seq.  Also, the message should suggest setting it to False, since True is the default.


AndreasHeger:
Thanks!

PR #317, opened 2016-07-24
Fix spelling/grammar issues found using Debian QA tools

afif-elghraoui:
The package linter picked these up and the notices don't go away without being fixed :)


AndreasHeger:
Thanks!

PR #319, opened 2016-07-28
Fix docstring in refskip

kyleabeauchamp:
There is no docstring on the `is_refskip` property on pileup reads.  I added the following one, but I am not 100% sure if my interpretation is correct:

"1 iff the base on the padded read is skipped in the reference, i.e. an insertion"

### DIFF
--- a/pysam/calignedsegment.pyx
+++ b/pysam/calignedsegment.pyx
@@ -2448,7 +2448,7 @@ cdef class PileupRead:
             return self._is_tail
 
     property is_refskip:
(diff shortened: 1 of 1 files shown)


jmarshall:
Close, but it's specifically cigar `N` rather than an insertion.  See documentation additions in samtools/htslib#226, which you might want to copy for `is_refskip` here.

kyleabeauchamp:
Thanks, I copied the docstring from HTSLib.

AndreasHeger:
Thanks

PR #322, opened 2016-08-03
Introducing find_introns

TyberiusPrime:
Add AligmentFile.find_introns(read_iterator): returns {(start, stop): read count} for all 'skipped' regions in the reads returned by read_iterator.

I find this a very nice addition to count_coverage to see what (if any) introns your reads support.

The interface right now is iterator based - I think, if there's no performance hit, that this make a much cleaner interface and implementation than passing reference, start, end, region, and a possible filter function to the various read counting functions. They all end up calling fetch anyhow, and this is more universal.


AndreasHeger:
Thanks, this is useful. I like the iterator interface.

I don't know about performance, something to be looked at later. The loop might execute quicker
by specifying

```python 
cdef AlignedSegment r
```

PR #333, opened 2016-08-26
Update calignedsegment.pyx docstring for query_length

amblina:
Quick change to the query_length method docstring as infer_query_length is in AlignedSegment not AlignmentFile plus errant '.'


AndreasHeger:
Thanks!

Issue #332, opened 2016-08-26
Hardclipped bases in cigar ignored by pysam.AlignedSegment.infer_query_length()


amblina:
The [ `pysam.AlignedSegment.infer_query_length`](http://pysam.readthedocs.io/en/latest/api.html#pysam.AlignedSegment.infer_query_length) method ignores hardclipped bases.  The method states that it gets the:

> inferred read length from CIGAR string.

this is in contrast with the [`pysam.AlignedSegment.query_length`](http://pysam.readthedocs.io/en/latest/api.html#pysam.AlignedSegment.query_length)

> This value corresponds to the length of the sequence supplied in the BAM/SAM file. The length of a query is 0 if there is no sequence in the BAM/SAM file. In those cases, the read length can be inferred  from the CIGAR alignment, see `pysam.AlignmentFile.infer_query_length()`.

With a cigar value of `58H150M` which represents an alignment where 58 bases have been hardclipped (not present in the alignment or SEQ section) and 150 matches. The SEQ section has 150 bases.  If `query_length=150` and `infer_query_length()=150`, however this should be: `infer_query_length()=208` by my understanding of the docs.  Am I missing something / is this intentional or just something to be corrected at some point?

Example of this can be found and run here: https://gist.github.com/amblina/795d30b5e04f87090d61739cb1846c1b 

Output:

```
@HD VN:1.3  SO:unsorted
@SQ SN:Test_reference   LN:200
query1  0   Test_reference  51  255 58H150M *   0   0   TCTCCGACTGTTCCCAAATTGTAACTTATCGTTCCGTGAAGGCCAGAGTCACTTCCCGGCCCTTTCCATGCGCGCACCATACCCTCCTAGTTCCCCGGTTATCTTTCCGAAGTGGGAGTGAGCGAACCTCCGTTTACGTCTTGTTACCAA  *   NM:i:1  AS:i:148

Infered length of query: 150
SEQ field length: 150
CIGAR field: 58H150M, length: 208
Actual length of original query: 208
```

AndreasHeger:
Merged and closed.

AndreasHeger:
Hi @amblina, just thought about it a bit more. I think the current behaviour was intentional to ensure consistency for infer_query_length when setting always=True and always=False to avoid the following situation:

```
>>> read.infer_query_length(True)
40
>>> read.infer_query_length(False)
30
```

[...]

PR #338, opened 2016-09-07
Fix for #332: Hardclipped bases in cigar ignored by pysam.AlignedSegment.infer_query_length()

amblina:
https://github.com/pysam-developers/pysam/issues/332
BAM_CHARD_CLIP ops in cigar are now used as part of calculation to infer original query length.  I have also added a test for infer_query_length().


AndreasHeger:
I can't recall any reason why hard-clipped bases should be excluded from the query length. Many thanks for the PR!

PR #345, opened 2016-09-16
Install fails on Ubuntu 14.04 when using pysam as requirement for other module

K377U:
When running make print-config first row contains

make[1]: Entering directory `/tmp/pip-build-eXV8ja/pysam/htslib'

This makes current code break.


AndreasHeger:
Many thanks!

Issue #360, opened 2016-10-26
pysam breaks with Cython 0.25


bioinformed:
Cython 0.25 has updated `Cython.Distutils.build_ext`, which has broken pysam with the following error on OSX, Linux and likely all OSes and Python versions:

```
    File "/usr/local/lib/python2.7/dist-packages/pysam/__init__.py", line 5, in <module>
    from pysam.libchtslib import *
ImportError: dynamic module does not define init function (initlibchtslib)
```

The problem is that the source is `pysam/chtslib.pyx`, yet the extension is named `pysam.libchtslib`.  For reasons I don't completely understand, the module init function behaves differently even though both versions of Cython create the same module init function `initchtslib` and under Cython 0.25 the above error results.  Or more importantly: I don't understand how this worked at all under previous versions of Cython.

This problem is a _big deal_, since pysam is going to break as soon as Cython is upgraded and it is rebuilt.  I have a PR ready for a fix that has been tested on OS X and will shortly be tested on Linux for Python versions 2.7.12, 3.4.5 and 3.5.2 and Cython versions 0.24.1 and 0.25.  Rather than renaming the chtslib.{pyx,pxd} source files to libchtslib.{pyx,pxd}, I've reworked how internal modules link to the chtslib shared library.  Maybe this is the right thing to do -- it certainly makes some of the internal logic in setup.py less hairy -- but maybe it isn't.  The consequence is that external modules that need to link against pysam will have to change yet again.

AndreasHeger:
Many thanks.  I had on my todo-list for the next "breaking" change to rename all the extension modules to ones starting with 'lib'. This makes linking and packaging much easier. Is now a good time for this and would this also address the cython issue? See #316.

Currently traveling, will get back more fully.

bagavi:
I get the following error after merging the above patch https://github.com/bioinformed/pysam/commit/6473cae1d51e3ebc528efa35ff727b923905f50a

----> 1 import pysam
/home/vivek/softwares/pysam/pysam/__init__.py in <module>()
      3 import sysconfig
      4 
----> 5 from pysam.chtslib import *
      6 from pysam.cutils import *
      7 import pysam.cutils as cutils

ImportError: No module named chtslib

bioinformed:
@bagavi: What OS, compiler and Python version are you using?

bioinformed:
@AndreasHeger: If that is the best way to proceed, I will implement that in my tree.

bagavi:
[...]

Issue #316, opened 2016-07-22
Use lib prefix for shared libraries


AndreasHeger:
... will making linking so much easier

PR #361, opened 2016-10-26
Fix Issue #360: pysam breaks with Cython 0.25

bioinformed:
Fix Issue #360.  May not be the ideal fix, because pysam.libchtslib is now called pysam.chtslib.  Fix tested on OS X and Linux for Python versions 2.7.12, 3.4.5 and 3.5.2 and Cython versions 0.24.1 and 0.25.

### DIFF
--- a/cy_build.py
+++ b/cy_build.py
@@ -87,6 +87,7 @@ class cy_build_ext(build_ext):
             if not ext.extra_link_args:
                 ext.extra_link_args = []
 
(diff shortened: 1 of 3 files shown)


bioinformed:
Strike that.  Linux builds are failing on TravisCI.  Investigating.

bioinformed:
Working on a different solution that renames all extension modules (per Issue #316)

Issue #260, opened 2016-03-24
Installation on Python 3 misses Cython dependency (0.9.0 regression)


martijnvermaat:
Since release 0.9.0, `pip install pysam` on Python 3 fails because Cython is missing:

```
$ virtualenv -p `which python3.5` py35
$ . py35/bin/activate
$ pip install pysam
Collecting pysam
  Using cached pysam-0.9.0.tar.gz
    Complete output from command python setup.py egg_info:
    checking for gcc... gcc
    checking whether the C compiler works... yes
...
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-8096gdno/pysam/setup.py", line 290, in <module>
        .format(fn))
    ValueError: no cython installed, but can not find pysam/chtslib.c.Make sure that cython is installed when building from the repository
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-8096gdno/pysam
```

Release 0.8.4 was good.

I tested Python 3.3, 3.4, and 3.5, they all fail in the same way. Python 2.7 seems to be good though, I'm not sure why.

It looks like the `requires` argument to setup is ignored (it lists `cython (>=0.21)`). I have no idea what causes this, but there have been quite some changes to `setup.py` since 0.8.4.

zyxue:
I looked a bit into this issue, here is what I found. With python-2.7 and pip-8.1.1, `pip install pysam` finishes successfully, but it doesn't install cython, either. Instead, it uses the `pysam/chtslib.c` that comes with `pysam-0.9.0.tar.gz`. You can check the stuffs inside the tar ball after `pip download pysam`.

The steps are shown as below
```
pip --version
pip 8.1.1 from /path/to/venv2/local/lib/python2.7/site-packages (python 2.7)
```
```
pip download pysam
Collecting pysam
  Using cached pysam-0.9.0.tar.gz
  Saved ./pysam-0.9.0.tar.gz
Successfully downloaded pysam
```
However, if I repeat the above process with Python-3.5 (I used Python-3.5 from Anaconda3, so the virtual environment layout is a bit different from python-2.7)
```
pip --version
pip 8.1.1 from /path/to/venv3/lib/python3.5/site-packages (python 3.5)
```
```
pip download pysam
Collecting pysam
  Using cached pysam-0.9.0.tar.gz
  Saved ./pysam-0.9.0.tar.gz
    Complete output from command python setup.py egg_info:
[...]

PR #365, opened 2016-10-27
Issue 316 360 rename shared libs

bioinformed:
Implement issue #316 to change all Cython modules to have a `lib` prefix.  This is an alternative way to correct issue #360.  Tested on Linux and OSX with Python 2.7.12, 3.4.5, and 3.5.2 with Cython 0.24.1 and 0.25.


kyleabeauchamp:
This looks like it cleans up a lot of redundant special-case code, which IMHO is a big win for maintainability.

bioinformed:
@AndreasHeger: The code looks good and the test suite passes, but I don't understand the TravisCI failures.  Can you please take a look when you get a chance?

kyleabeauchamp:
OK, I wonder if this is the issue: https://github.com/pysam-developers/pysam/blob/master/run_tests_travis.sh#L135

I think in the past we had cython as an optional dependency, so we had a manual test of with and without cython installed in the CI script.  AFAIK now that might be deprecated.  Correct me if I'm wrong.

kyleabeauchamp:
Actually that doesn't seem like it.  Hmm

kyleabeauchamp:
Related to #260 perhaps?

kyleabeauchamp:
I stared at the setup script some more and it indeed strange that "pysam/libchtslib.pyx" is claimed to be missing and tripping the "no cython" code path.  

Do we need to support both cython and nocython builds?  IMHO it would be great to reduce the number of code paths in `setup.py` and in the CI tests.  One step in that direction would be to make cython a hard dependency and simplify the build.

AndreasHeger:
Thanks both, I will have a look now.

AndreasHeger:
Hi @bioinformed and @kyleabeauchamp. The issue was that the travis script ran:

```
python setup.py sdist
...
pip install pysam-xxx.tar.gz
```

However, the dist command does not cythonize .pyx files into .c. The fix was to run:

```
python setup.py build sdist
...
pip install pysam-xxx.tar.gz
```

I am not sure why it used to work. Could be that .c files were accidentally part of the repos or this might be change with the cython 0.25 builder?

AndreasHeger:
@bioinformed, many thanks for the refactoring!

Issue #344, opened 2016-09-15
pysam wheels not available at www.cgat.org?


cgates:
Until today there was a pysam wheelhouse at:
https://www.cgat.org/downloads/public/pysam/wheels/
(The wheels make using TravisCI much nicer.)
The cgat.org site appears to be down: is that no longer supported?
If not, is there a better source for pysam wheels?

Thanks!
cg

AndreasHeger:
Hi Cg,

we are moving the server to a new network (new IP adresses, etc),
but it should reappear again as www.cgat.org.

It should be back up soon, I will ask.

Best wishes,
Andreas

cgates:
All is well.  Thanks!!!

graingert:
@AndreasHeger can you publish manylinux1 wheels on PyPI instead?

AndreasHeger:
Ok, will do.

AndreasHeger:
Pysam wheels now on pypi.

PR #374, opened 2016-11-15
Script for building manylinux1 wheels

marcelm:
As discussed in issue #344, it would be great if wheels for pysam were available. Pysam is quite a heavy dependency, and it is one of those packages for which offering wheels would therefore really pay off by reducing installation time.

This script is based on the [manylinux1 demo](https://github.com/pypa/python-manylinux-demo). It uses docker to build binary, manylinux1 wheels. It is meant to be run locally for now. The wheels are put into a `wheelhouse/` directory, and someone with access to the PyPI account then needs to upload them (I guess with `twine upload`).

The only limitation is that HTTPS and S3 URLs are not supported because this would mean that OpenSSL needs to be bundled into the wheel, which in turn would make pysam developers responsible for updating the wheels when there is a security problem. This would probably need to be communicated.

### DIFF
--- a/buildwheels.sh
+++ b/buildwheels.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+#
+# Build manylinux1 wheels for pysam. Based on the example at
+# <https://github.com/pypa/python-manylinux-demo>
+#
+# It is best to run this in a fresh clone of the repository!
+#
+# Run this within the repository root:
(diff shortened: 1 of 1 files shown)


AndreasHeger:
Great, many thanks!

I will play around with the script to get the wheels up. Currently busy at work, aiming for a new release in three weeks (early Dec).

marcelm:
Great, thanks for merging!

I’m following the development around the manylinux1 standard and will let you know when there is a solution for the OpenSSL thing. I don’t see it as a blocker, though.

Issue #349, opened 2016-09-29
pip install - ValueError: dictionary update sequence element #0 has length 3; 2 is required


tseemann:
Installed via pip3 in virtualenv:

```
  Running setup.py (path:/tmp/pip-kb5lfk5k-build/setup.py) egg_info for package from file:///tmp/ariba--pysam-20160930-1
20692-1c7jwaf/pysam-0.9.1.4
    Running command python setup.py egg_info
    checking for gcc... /home/linuxbrew/.linuxbrew/bin/gcc-5 
checking whether the C compiler works... yes            
<snip>
config.status: creating config.mk                                                                                      
creating config.h
    # pysam: cython is available - using cythonize if necessary
    # pysam: htslib mode is shared
    # pysam: HTSLIB_CONFIGURE_OPTIONS=None
    # pysam: htslib configure options: --enable-libcurl
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-kb5lfk5k-build/setup.py", line 209, in <module>
             htslib_make_options = run_make_print_config()
      File "/tmp/pip-kb5lfk5k-build/setup.py", line 68, in run_make_print_config
      for line in stdout.splitlines()])
    ValueError: dictionary update sequence element #0 has length 3; 2 is required
Cleaning up...
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-kb5lfk5k-build/
```

EricDeveaud:
this is due to the call to subprocess.check_output(["make", "print-config"])
line 69 in setup.py
when python setup.py is called by make, output of make print-config is the following
```
make[1]: Entering directory `/tmp/pysam-0.9.1.4/htslib'
LDFLAGS =
LIBHTS_OBJS = kfunc.o knetfile.o kstring.o bgzf.o faidx.o hfile.o hfile_net.o hts.o md5.o regidx.o sam.o synced_bcf_reader.o vcf_sweep.o tbx.o vcf.o vcfutils.o cram/cram_codecs.o cram/cram_decode.o cram/cram_encode.o cram/cram_external.o cram/cram_index.o cram/cram_io.o cram/cram_samtools.o cram/cram_stats.o cram/files.o cram/mFILE.o cram/open_trace_file.o cram/pooled_alloc.o cram/rANS_static.o cram/sam_header.o cram/string_alloc.o cram/thread_pool.o cram/vlen.o cram/zfio.o hfile_libcurl.o
[...]

PR #379, opened 2016-12-01
Simple fix to ValueError issue during CI build

standage:
Simple fix based on @EricDeveaud's comment on #349. I was getting the same error when trying to install with `pip install pysam` on Travis CI. Now it works fine with `pip install https://github.com/standage/pysam.git`. Obviously, it would be nice if this got fixed in the next release so we could all go back to using the canonical pip version!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,7 +60,7 @@ def run_configure(option):
 
 
 def run_make_print_config():
-    stdout = subprocess.check_output(["make", "print-config"])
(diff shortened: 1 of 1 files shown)


jmarshall:
Is this not already fixed by #345, which is a better approach?

standage:
Yes, #345 fixes my issue. I should have thought to try pip installing the latest master, which indeed works. When you cut a new release I'll go back to installing from PyPI.

Issue #371, opened 2016-11-04
INFO Field Parsing String -127


paudano:
When parsing INFO, a field with -127 becomes an empty tuple in `pysam.cbcf.VariantRecord.info`

Python: 3.5.2
Pysam: 0.9.1.4
* Also replicated in Pysam 0.8.4 on Python 2.7.12

Code to replicate (see "short.vcf" below).

```$ python
Python 3.5.2 (default, Jun 28 2016, 08:46:01) 
[GCC 6.1.1 20160602] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> import pysam
>>> 
>>> vcf = pysam.VariantFile('short.vcf')
>>> 
>>> record = next(vcf)
>>> record.info['SVLEN']
()
>>> 
>>> record = next(vcf)
>>> record.info['SVLEN']
-126
>>> 
>>> record = next(vcf)
>>> record.info['SVLEN']
-128
>>> 
>>> 
>>> sys.version_info
sys.version_info(major=3, minor=5, micro=2, releaselevel='final', serial=0)
>>> 
>>> pysam.__version__
'0.9.1.4'
>>>
```

short.vcf
```
##fileformat=VCFv4.2
##contig=<ID=chr1,length=248956422>
##INFO=<ID=END,Number=1,Type=Integer,Description="Reference end position of the variant">
##INFO=<ID=SVTYPE,Number=1,Type=String,Description="Type of structural variant">
##INFO=<ID=SVLEN,Number=1,Type=Integer,Description="Difference in length between the REF and ALT alleles">
##FORMAT=<ID=GT,Number=1,Type=Integer,Description="Genotype">
#CHROM  POS     ID      REF     ALT     QUAL    FILTER  INFO    FORMAT  SAMPLE
chr1	100000	TEST-127	C	<DEL>	100	PASS	END=100126;SVTYPE=DEL;SVLEN=-127	GT:CS:CD	1/1
chr1	200000	TEST-126	C	<DEL>	100	PASS	END=200125;SVTYPE=DEL;SVLEN=-126	GT:CS:CD	1/1
chr1	300000	TEST-128	C	<DEL>	100	PASS	END=300127;SVTYPE=DEL;SVLEN=-128	GT:CS:CD	1/1
```

bioinformed:
Thanks for the issue report!  The problem is clear and I'll hopefully have it triaged and solved quickly.

bioinformed:
[...]

Issue #355, opened 2016-10-14
Implement streams using file descriptors


hzpc-joostk:
Hi everyone!

I am working on streaming pysam applications for working with both SAM/BAM and VCF/BCF data. Streams from/to stdin/stdout could be parsed/written by passing `"-"` as the filename argument, which is very neath! However, sometimes I would like to pre-parse the data stream using a dedicated samtools/bcftools command. 

Within Python I redirect Python's stdin to a sub-process utilizing the required command, then redirecting its output to Python's stdin using `os.dup2`. This way the pysam engine could get the data stream from the sub-process without the need of an intermediate file. Here's a code snippet of this approach (Python 3):

``` python
...

args = ['bcftools',            # The bcftools executable
        'view',                # The bcftools command
        '--output-type', 'u',  # Output uncompressed BCF to my STDIN
        '--no-update'          # do not (re)calculate INFO fields
    ] + args                   # Additional arguments

# Create an OS PIPE for (IPC)
parent_read, child_write = os.pipe()

# Launch the sub-process
try:
    proc = subprocess.Popen(
        args,
        stdin=sys.stdin,      # Read from my STDIN
        stdout=child_write    # PIPE the childs STDOUT to the IPC PIPE
    )
finally:
    # be sure the FD is closed no matter what
    os.close(child_write)

# Duplicate the child's STDOUT to my STDIN (not affecting the child's STDIN)
# Note: do this AFTER starting the child!
os.dup2(parent_read, sys.stdin.fileno())

# pysam.VariantFile can now read the child's STDOUT via my STDIN
with proc, pysam.VariantFile('-', 'r') as vcf:
    for rec in vcf:
        do_something(rec)

...
```

This snippet was based on a code snippet from Programming Python, 4th edition, chapter 5, page 229 (Interprocess Communication). Again as seen it creates a child process running the `bcftools view`, reading Python's `sys.stdin`, writing to a pipe which is duplicated (`os.dup2`) onto Python's stdin. 

The additional arguments for the sub-process could be a region, sample subset, filters or even a filename when not streaming stdin.

Writing a file stdout (fd=2) is possible by passing `"-"`, but writing to multiple streams is not possible now. It would be great if a file descriptor could be passed as a filename, so this redirecting/duplication of file descriptors becomes unnecessary. 

Thanks for your thoughts and considerations!

Cheers, Joost

hzpc-joostk:
[...]

PR #385, opened 2016-12-11
More bcf fixes

bioinformed:
* Issue #371: do not interpret vector end codes for scalar values	
* Issue #355: Implement streams using file descriptors for VariantFile
* Fix Python 3 build on Darwin
* Refactor common functionality from AlignmentFile and VariantFile into a common HTSFile base class


bioinformed:
@AndreasHeger: You may want to review this PR, since it adds a common base class `HTSFile` for `AlignmentFile` and  `VariantFile` and exposes more `htsFormat` information.

kyleabeauchamp:
Should this base class be used for TabixFile as well?  It seems to have at least some of the same fields (e.g. filename, index).

bioinformed:
Good point, @kyleabeauchamp!   `TabixFile` now inherits from `HTSFile` too.

pysam/libctabix.pyx
```diff
@@ -72,7 +72,7 @@ from pysam.libchtslib cimport htsFile, hts_open, hts_close, HTS_IDX_START,\
     BGZF, bgzf_open, bgzf_dopen, bgzf_close, bgzf_write, \
     tbx_index_build, tbx_index_load, tbx_itr_queryi, tbx_itr_querys, \
     tbx_conf_t, tbx_seqnames, tbx_itr_next, tbx_itr_destroy, \
-    tbx_destroy, hisremote
+    tbx_destroy, hisremote, region_list
```
kyleabeauchamp: What's the deal with `region_list`?  It's newly imported here but commented out later.

AndreasHeger:
Thanks, having HTSFile is clearly the right thing to do.

PR #389, opened 2016-12-26
setup.py: remove redundant setting of HTSLIB_MODE

afif-elghraoui:
It's already initialized earlier in the script.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -145,15 +145,11 @@ try:
     HAVE_CYTHON = True
     print ("# pysam: cython is available - using cythonize if necessary")
     source_pattern = "pysam/libc%s.pyx"
-    if HTSLIB_MODE != "external":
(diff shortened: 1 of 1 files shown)


afif-elghraoui:
This code was there before you added the line:
```python
HTSLIB_MODE = os.environ.get("HTSLIB_MODE", "shared")
```
earlier on in the script.  However, the code that this patch removes makes it impossible to ever use HTSLIB_MODE=separate, because it would get overridden to "shared" no matter what (in both the try and except clauses). I don't know if you had intended something different.

Just an additional note-- we don't need this patch for Debian (we're using "external" for this mode), but since I had added part of the code here, I felt like I should clean it up since it looks like it's now obsolete.

AndreasHeger:
Thanks! I will need to go through setup.py and simplify.

PR #393, opened 2017-01-18
Ah add file descriptor support to alignment file

AndreasHeger:
Addresses #355 for AlignmentFile objects

### DIFF
--- a/pysam/libcalignmentfile.pyx
+++ b/pysam/libcalignmentfile.pyx
@@ -58,6 +58,9 @@ import re
 import warnings
 import array
 
+from libc.errno  cimport errno, EPIPE
+from libc.string cimport strcmp, strpbrk, strerror
+from posix.fcntl cimport fcntl, F_GETFD
 from cpython cimport array as c_array
 from cpython.version cimport PY_MAJOR_VERSION
 
@@ -404,6 +407,10 @@ cdef class AlignmentFile(HTSFile):
         if referencelengths is not None:
             reference_lengths = referencelengths
 
+        # close a previously opened file
+        if self.is_open:
+            self.close()
+
         # autodetection for read
         if mode is None:
             mode = "r"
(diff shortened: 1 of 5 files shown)


AndreasHeger:
... still requires a test.

pysam/libcalignmentfile.pyx
```diff
@@ -413,38 +420,41 @@ cdef class AlignmentFile(HTSFile):
...
+            self.is_remote = False
+            self.is_stream = True
+        # reading from a File object or other object with fileno
         elif hasattr(filepath_or_object, "fileno"):
-            filename = filepath_or_object.name
             if filepath_or_object.closed:
                 raise ValueError('I/O operation on closed file')
+            self.filename = filepath_or_object
+            # .name can be TextIOWrapper
+            filename = encode_filename(str(filepath_or_object.name))
```
bioinformed: What is `name` doesn't exist (i.e. `AttributeError` is raised)?

pysam/libchtslib.pyx
```diff
@@ -244,3 +250,12 @@ cdef class HTSFile(object):
             cfilename = filename
             with nogil:
                 return hts_hopen(hfile, cfilename, cmode)
+
+    def _exists(self):
+        """return False iff file is local, a file and exists.
+        """
+        if (isinstance(self.filename, (str, bytes)) and
```
bioinformed: better as?
```python
    return (not isinstance(self.filename, (str, bytes))
            or self.filename == b'-' or self.is_remote
            or os.path.exists(self.filename))
```

pysam/libcalignmentfile.pyx
```diff
@@ -501,15 +511,17 @@ cdef class AlignmentFile(HTSFile):
...
-            if hasattr(filepath_or_object, "fileno"):
-                fp = hdopen(filepath_or_object.fileno(), cmode)
-                with nogil:
-                    self.htsfile = hts_hopen(fp, cfilename, cmode)
-            else:
-                with nogil:
-                    self.htsfile = hts_open(cfilename, cmode)
+            self.htsfile = self._open_htsfile()
+
+            # # open file (hts_open is synonym with sam_open)
```
bioinformed: This block no longer needed.

pysam/libchtslib.pyx
```diff
@@ -222,6 +226,8 @@ cdef class HTSFile(object):
                 fd = self.filename
             else:
                 fd = self.filename.fileno()
+                
+            dup_fd = dup(fd)
```
bioinformed: Should we support passing a `close_file=True` parameter to `HTSFile` objects to control whether the `fd` should be `dup`ed?

pysam/libcalignmentfile.pyx
AndreasHegerGenomics: Right, will catch this.

pysam/libcalignmentfile.pyx
AndreasHeger: Thanks, removed!

pysam/libchtslib.pyx
AndreasHeger: Yes, we can. Probably close_file = False and dup should be default? 
Otherwise I can see no way to prevent double closing errors when used with File objects.

pysam/libchtslib.pyx
AndreasHeger: Done!

pysam/libchtslib.pyx
AndreasHeger: ... on second thoughts, I wonder if it might get a bit confusing for the average user as they wonder if they should or should not call AlignmentFile.close(), etc.
Maybe being explicit is better such as "own_filehandle" or "duplicate_filehandle" for advanced user is clearer?

bioinformed: I agree-- I like `dup_fd` or `duplicate_filehandle`.

pysam/libcalignmentfile.pyx
```diff
@@ -58,6 +58,9 @@ import re
 import warnings
 import array
 
+from libc.errno  cimport errno, EPIPE
+from libc.string cimport strcmp, strpbrk, strerror
+from posix.fcntl cimport fcntl, F_GETFD
```
bioinformed: Not used?

AndreasHeger: Thanks, removed (fcntl)

pysam/libchtslib.pyx
AndreasHeger: Thanks, I will add this and release tomorrow.

PR #394, opened 2017-01-19
{AH} enable py3.6

AndreasHeger:


kyleabeauchamp:
FWIW, the bioconda infrastruction hasn't enabled 3.6 just yet, but it's ongoing:

https://github.com/bioconda/bioconda-recipes/pull/3338

AndreasHeger:
Thanks. Interestingly, it worked. bcftools, etc. were picked up, but compiled against python3.6.

kyleabeauchamp:
Yep, there should be no problem building pysam because the bioconda dependencies are all non-python libraries.

The only possible issue is that we'll need bioconda to enable 3.6 so that they can build the pysam py3.6 package once you finalize the release.

PR #401, opened 2017-02-06
Ah runpat hissues

AndreasHeger:


bioinformed:
I'll test on Linux and OSX and let you know if I can provoke any failures.

AndreasHeger:
seems ok, I will merge for now.

PR #408, opened 2017-02-15
Attempt to fix #406

### LINKED ISSUE #406: set_tags() can't handle byte-array auxiliary tags
This is originally from [biostars](https://www.biostars.org/p/236819/). If you try to `set_tags()` using a byte array then the following error is produced:

      File "./foo.py", line 9, in <module>
        read.set_tags(tags)
      File "pysam/libcalignedsegment.pyx", line 2113, in pysam.libcalignedsegment.AlignedSegment.set_tags (pysam/libcalignedsegment.c:23066)
      File "pysam/libcalignedsegment.pyx", line 295, in pysam.libcalignedsegment.packTags (pysam/libcalignedsegment.c:6217)
    ValueError: invalid value type 'B' (<type 'str'>)

I'm attaching a very very small BAM file (it and the index are in the zip file) that can produce this error with the following code:

    #!/usr/bin/env python
    import pysam

    bam = pysam.AlignmentFile("small.bam")
    for read in bam.fetch():
        tags = read.get_tags(with_value_type=True)
        read.set_tags(tags)

I've tried earlier version (back to 0.8.4) and none of these are able to handle this sort of situation.

[example.zip](https://github.com/pysam-developers/pysam/files/777372/example.zip)

dpryan79:
This seems to fix the issue I reported in #406. In short, for arrays the `valuetype` is always going to be set to `B`, which won't be appropriate for the type of the array. This code changes this to the type of the array or `None` if that's not a known type (e.g., if a known type ends up getting stored as something else...the ZM tag in my original example in #406 is such a case) then this is changed to `None` and generated as needed. It's likely that this means that some of the other code in `packTags()` is no longer needed.

### DIFF
--- a/pysam/libcalignedsegment.pyx
+++ b/pysam/libcalignedsegment.pyx
@@ -281,6 +281,9 @@ cdef inline packTags(tags):
                          len(value)] + list(value))
 
         elif isinstance(value, array.array):
+            valuetype = value.typecode
+            if valuetype not in datatype2format:
+                valuetype = None
             # binary tags from arrays
             if valuetype is None:
                 array_typecode = map_typecode_python_to_htslib(ord(value.typecode))


AndreasHeger:
Great, thanks!

Issue #417, opened 2017-03-16
pip install fails for python 3.3.6 (using venv)


bamyasi:
Hi,

I am trying to install pysam 0.10.0 on our Linux cluster. No matter which combination of HTSLIB_MODE and HTSLIB_SOURCE I tried, 'pip install pysam' would always fail with the following error from the linker:

/usr/bin/ld: cannot find -lchtslib

This is python 3.3.6 and pip 9.0.1 run in a virtual environment (created via venv). I have also tried 'python setup.py install' method but it fails with the same error.

Best,
Ivan

AndreasHeger:
Sorry, but we don't support python 3.3 any more. I don't recall the exact reason, but I think it was due to incompatibilities with linking/building to later python 3 versions. Is it possible to upgrade?

PR #418, opened 2017-03-17
update setup.py to use correct sysconfig variable for py3

abjonnes:
This addresses #417. The `SOABI` and `EXT_SUFFIX` variables were introduced in py3.2 with [PEP 3149](https://www.python.org/dev/peps/pep-3149/), which apparently broke some extensions that use `SO`. This problem was fixed and `SO` was deprecated in py3.4. See [issue16754](https://bugs.python.org/issue16754) for some more discussion.


abjonnes:
Apparently extension suffices on OSX did not use ABI until py3.5, so we should use `EXT_SUFFIX` directly in all cases, falling back to `SO` if not available, since this is what is used by `distutils` internally.

AndreasHeger:
Thanks, apologies for the delay in getting back, this makes sense.

PR #428, opened 2017-04-06
Htslib 1.4

AndreasHeger:
Incorporate htslib/bcftool/ssamtools 1.4


kyleabeauchamp:
FWIW tests seem almost all good on my Ubuntu 16.04 + anaconda3.6 environment, modulo what is likely a missing file or something:

```

======================================================================
ERROR: testFTPView (faidx_test.TestRemoteFileFTP)
----------------------------------------------------------------------
  File "~/src/software/pysam/tests/faidx_test.py", line 229, in testFTPView
    with pysam.Fastafile(self.url) as f:
  File "pysam/libcfaidx.pyx", line 114, in pysam.libcfaidx.FastaFile.__cinit__ (pysam/libcfaidx.c:2351)
  File "pysam/libcfaidx.pyx", line 164, in pysam.libcfaidx.FastaFile._open (pysam/libcfaidx.c:3540)
ValueError: could not locate index file GRCh38_full_analysis_set_plus_decoy_hla.fa.fai
```

AndreasHeger:
Thanks. This is remote file access and the URL is valid, the error might a bit more than missing file. I will check.

kyleabeauchamp:
Any thoughts on syncing the pysam versioning to match HTSLib and friends?  It might help reduce confusion when people have different versions sitting around.

kyleabeauchamp:
E.g. if everything was 1.4 it might be less confusing that HTSLib 1.4 and pysam 0.11

AndreasHeger:
@kyleabeauchamp, good point. @bioinformed and I discussed this before (off-line) and our thoughts were that pysam is not quite >1.0 yet, the VariantFile API is not yet fully settled. But that aside, it might also be difficult to tie both too closely together if we want to use semantic versioning for pysam to indicate breaking changes to the pysam API while the underlying htslib remains unchanged. Do you know of a good versioning scheme that would allow this?

kyleabeauchamp:
I don't know of a good alternative versioning scheme.  I imagine there might be Boost "add-on-libraries" that are independent of Boost but whose versioning is tied to the Boost version--however I can't think of any offhand.

I agree that fixing to HTSLib would reduce our ability to separately implement semantic versioning.  I think your rationale is a convincing reason to stick with status quo for a while.

Issue #433, opened 2017-04-10
Several errors due to bug in getQueryEnd


bioinformed:
When a sequence is not present in an `AlignedSegment`, the `getQueryEnd` function took the result from `calculateQueryLength` as the length of the query sequence and subtracted any trailing soft-clipping.  This would be fine except that `calculateQueryLength` also includes hard-clipped bases.

`getQueryEnd` and `query_alignment_length` now return incorrect values when a sequence is not present and CIGAR includes hard clipping.  This was corrected by updating the implementation of `getQueryEnd` to compute the proper value without calling `calculateQueryLength`.

`calculateQueryLength` is still used to compute `infer_query_length`, which will return different results depending on whether the observed query length is used or not.  Rather than an *always* flag, it seems like it should have a flag for *include_hard_clipping*, since that is the source of the difference between using the sequence or the CIGAR only in properly formed SAM files.

PR #434, opened 2017-04-10
[#433] correct getQueryEnd

bioinformed:
See #433


bioinformed:
@AndreasHeger: Please review.  

I'm pretty confident about the fix itself.  However in addressing it, I've decided that `infer_query_length()` is not sensible.  I recommend getting rid of the *always* flag (if possible) and adding a flag to control whether the result includes hard-clipped bases or not.

AndreasHeger:
Thanks, looks much better. I will open a separate issue about getting rid of the always flag.

PR #448, opened 2017-04-21
Fix #440

### LINKED ISSUE #440: infer_query_length() is missing the "always" keyword option.
As of the last release, the `infer_query_length()` function apparently lost the `always=` keyword argument. This is very problematic, since it breaks a lot of existing code (e.g., deepTools). I'm happy to make a PR to fix this if needed.

dpryan79:
This fixes the issue I reported in #440 with `infer_query_length()`. Old code specifying `always` will now use the new functions.

I'm getting a bug report every day or so now from people using deepTools that's due to this, so a quickish 0.11.1 release would be awesome :) If you want to pack anything else into it and need a hand then I'm happy to help.

### DIFF
--- a/pysam/libcalignedsegment.pyx
+++ b/pysam/libcalignedsegment.pyx
@@ -1457,14 +1457,19 @@ cdef class AlignedSegment:
 
         return result
 
-    def infer_query_length(self):
+    def infer_query_length(self, always=False):
         """infer query length from sequence or CIGAR alignment.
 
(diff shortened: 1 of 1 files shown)


AndreasHeger:
Thanks, looks good. I will add a deprecation statement to the docs and then release.

dpryan79:
Awesome, thanks @AndreasHeger!

Issue #473, opened 2017-06-02
Add example of writing fastq


kyleabeauchamp:
#471 asked for fastq writing capabilities.  I think we already have this, but it's a bit hard to find.  Basically, the `__str__()` member function FastqProxy does what we need for writing, so we might as well add an example of this.

```
for line in fastq:
   fh.write(str(line))
```

The only other thing that might be desirable is to have a constructor for FastqProxy that creates an object from the name, seq, and qual strings.

AndreasHeger:
Good points, I will see if I can polish this part this weekend.

AndreasHeger:
Done, see 1c28fcb1277e16da31ca5af11fc18ad737dd862e

PR #474, opened 2017-06-02
Add example of fastq output

kyleabeauchamp:


kyleabeauchamp:
cc #473

kyleabeauchamp:
I didn't check that this formats OK in the output HTML due to some local build issues I was having.

AndreasHeger:
Thanks

PR #489, opened 2017-06-20
Use @loader_path for mac OS library paths

jvkersch:
This is somewhat of an experimental PR to solicit some feedback on an issue with building pysam eggs/wheels on mac OS. The issue is that the current configuration in [cy_build.py](https://github.com/pysam-developers/pysam/blob/d41a7c87d93f50277f4a2d1f0efc07fa19954963/cy_build.py#L80) will cause the absolute path to the interpreter to be used in the libraries' rpath, and this makes an egg/wheel unshareable between different machines, or even different Python installations on the same machine. 

To see this in more detail:
1) Build a wheel; `python setup.py bdist_wheel`
2) Unpack the egg; `cd dist && unzip pysam*.whl`
3) Inspect the rpath for any of the shared libraries (note the absolute path):
```
$ otool -l pysam/libcutils.*.so | grep -A2 LC_RPATH
          cmd LC_RPATH
      cmdsize 88
         path /Users/jvkersch/miniconda3/envs/pysam-env/lib/python3.6/site-packages (offset 12)
```

This PR removes the explicit setting of rpath, and changes the install_name for the shared libraries to something of the form `@loader_path/<library-name>`.  I was able to run the tests successfully on a clean install from wheels, and in a (different) environment with `python setup.py develop`.

Perhaps there's something I've overlooked here, please let me know if that's the case!

### DIFF
--- a/cy_build.py
+++ b/cy_build.py
@@ -60,14 +60,20 @@ class cy_build_ext(build_ext):
             ext.library_dirs.append(os.path.join(self.build_lib, "pysam"))
 
         if sys.platform == 'darwin':
+            # The idea is to give shared libraries an install name of the form
+            # `@rpath/<library-name.so>`, and to set the rpath equal to
+            # @loader_path. This will allow Python packages to find the library
+            # in the expected place, while still giving enough flexibility to
+            # external applications to link against the library.
             relative_module_path = ext.name.replace(".", os.sep) + get_config_vars()["SO"]
             library_path = os.path.join(
-                "@loader_path", os.path.basename(relative_module_path)
+                "@rpath", os.path.basename(relative_module_path)
             )
 
             if not ext.extra_link_args:
                 ext.extra_link_args = []
             ext.extra_link_args += ['-dynamiclib',
+                                    '-rpath', '@loader_path',
                                     '-Wl,-headerpad_max_install_names',
                                     '-Wl,-install_name,%s' % library_path,
                                     '-Wl,-x']
--- a/cy_build.py
+++ b/cy_build.py
@@ -61,25 +61,15 @@ class cy_build_ext(build_ext):
 
         if sys.platform == 'darwin':
             relative_module_path = ext.name.replace(".", os.sep) + get_config_vars()["SO"]
-
-            if "develop" in sys.argv or "test" in sys.argv:
-                # develop-mode and tests use local directory
-                pkg_root = os.path.dirname(__file__)
-                linker_path = os.path.join(pkg_root, relative_module_path)
-            elif "bdist_wheel" in sys.argv or is_pip_install():
-                # making a wheel, or pip is secretly involved
-                linker_path = os.path.join("@rpath", relative_module_path)
-            else:
-                # making an egg: `python setup.py install` default behavior
-                egg_name = '%s.egg' % self._get_egg_name()
-                linker_path = os.path.join("@rpath", egg_name, relative_module_path)
+            library_path = os.path.join(
+                "@loader_path", os.path.basename(relative_module_path)
+            )
 
             if not ext.extra_link_args:
                 ext.extra_link_args = []
             ext.extra_link_args += ['-dynamiclib',
-                                    '-rpath', get_python_lib(),
                                     '-Wl,-headerpad_max_install_names',
-                                    '-Wl,-install_name,%s' % linker_path,
+                                    '-Wl,-install_name,%s' % library_path,
                                     '-Wl,-x']
         else:
             if not ext.extra_link_args:
--- a/run_tests_travis.sh
+++ b/run_tests_travis.sh
@@ -25,7 +25,7 @@ conda config --add channels defaults
 conda config --add channels r
 conda config --add channels bioconda
 
-conda install -y samtools bcftools htslib
+conda install -y "samtools=1.4" "bcftools=1.4" "htslib=1.4"
 
 # Need to make C compiler and linker use the anaconda includes and libraries:
 export PREFIX=~/miniconda3/


bioinformed:
Thanks for your work on this issue!  I've not tried using relative loader paths before.  I'll start testing the various cases, including install from the various modes and also linking from a dependent package.

bioinformed:
This patch fixes the initial installation problems on OS X, but only addresses linking from within the same directory.  pysam is frequently extended by Cython modules in other packages, which no longer works with this patch applied.  I think we're stuck with having non-movable binaries.

jvkersch:
Thanks @bioinformed for the quick review! Do you know of any specific packages that do what you described (i.e. with Cython modules that extend pysam)?

bioinformed:
[HTseq](https://github.com/simon-anders/htseq) is one of the more popular packages that links against pysam.  There are many others too, including substantial non-open source packages used by some leading NGS companies.  (I know because I've written a few of them and gotten support requests from many more.)

jvkersch:
Thanks. I'll look into this a little more, but this seems quite intricate...

bioinformed:
Also see: http://pysam.readthedocs.io/en/latest/usage.html#extending-pysam

This will give you a small test that is representative of how others are extending pysam.  It is a wonder this works on Linux out of the box.

jvkersch:
@bioinformed  I finally had some time to look into this properly. I think we can get relative paths to work in every situation (egg/wheel/local install), but I made a slight change to the setup: instead of using `@loader_path` directly in the install name, I reverted to using `@rpath` for the install name prefix, and setting `rpath` on the library to `@loader_path`. For packages that run inside Python, this is functionally identical to what I had before, but gives some flexibility to 3rd party applications that want to link against pysam: they can provide `-rpath` upon compilation to specify the pysam install location. Is that acceptable?

This dance with `@rpath` and `@loader_path` is similar to how Conda and Enthought EPD packages are built; see https://github.com/conda/conda-build/issues/279. The fix that I have here is also very similar to the existing [bioconda recipe for pysam](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/pysam/osx_rpath.patch), though that doesn't include the provision with `@rpath`.

To test this, I ran `python setup.py develop`, `python setup.py install`, and I installed from a platform wheel, each time in a clean environment. I then ran the tests, and I installed htseq and ran the tests for that package. In each case I got a clean test run (with one caveat, described below). I was also able to dynamically load the pysam libraries from an external C program, though I had to set `-rpath` when compiling.

The only issue that I had while testing was that there has been a new version of samtools/htslib/... in the last week or so. If this new version is installed in the environment, it interferes with the version of htslib which is vendorized with pysam. To work around this, I pinned the version numbers of samtools etc. to 1.4 in the travis setup script (2bf0869). I haven't properly investigated what is going on here.

jvkersch:
That said, I'd be happy to run additional tests!

bioinformed:
This looks extremely promising!  I'll start testing...

bioinformed:
Unfortunately this latest version is still failing for me.  I checked and HTseq seems to have dropped their use of linking against pysam, probably because it doesn't work reliably.  Here is one of my packages to use to test: https://github.com/bioinformed/vgraph

I think the solution will require setting an rpath relative to the Python binary or site-packages.  It is currently done via an absolute path, but I suspect it could be done via a relative path.

jvkersch:
> Unfortunately this latest version is still failing for me.

Can you tell me how you installed the package? Perhaps I can reproduce...

Thanks for pointing out the vgraph package; I installed that via pip in clean environment and ran the doctests, and everything passed.

bioinformed:
I've just retested from a new virtualenv and it is working now.  I need to go back and see if I can recreate that I did previous to cause the link failure.

jvkersch:
Could it be samtools/htslib/bcftools at version 1.5? That gave me a lot of grief...

bioinformed:
Nope-- I'm working on adding support for the new htslib/samtools/bcftools in another branch.  And I don't use the Travis script for local testing.

bioinformed:
@caethan: I know you've worked on the problem of pysam linking before.  Can you give this PR a try and see if it works for you?

bioinformed:
@AndreasHeger: I think @jvkersch may have fixed the linking problems on OS X.  Please review and merge if you're comfortable.

AndreasHeger:
thanks @jvkersch  and @bioinformed  - outside my expertise, sounds all good to me!

I will augment the travis test script to test linking agaiinst pysam libraries outside the pysam install directory.

jvkersch:
Thanks @bioinformed and @AndreasHeger ! Please ping me if there are any issues with this.

bioinformed:
Bad news.  vgraph uses the pysam pxd files, which works with this PR.  I have another package that links against functions defined in pysam and htslib, which does not.  Before reverting this PR, I'll see if I can track down the specific problem.

AndreasHeger:
Thanks. If you let me know how you link, I will add same tests.

jvkersch:
@bioinformed Thanks for the notification; I would like to look into this but I don't have much time this week. I will try to get to it over the weekend, or by next week at the latest.

bioinformed:
I'll create a dummy app, since I don't have permission to share this app.  However, here is how I link from setup.py:
```
ext_modules = [Extension('helix_app_covetous.libcovetous',
                        ['helix_app_covetous/libcovetous.pyx'],
                         include_dirs=pysam.get_include()+[np.get_include()],
                         extra_link_args=pysam.get_libraries(),
                         define_macros=pysam.get_defines(),
                         )]
```
(covetous is a coverage analysis app developed by Helix that extracts coverage information from BAMs, stores it in a highly compressed manner, and allows running queries on many thousands of samples very quickly.  It isn't open source yet, but that is planned when it is feature complete and stable.)

PR #502, opened 2017-07-10
Ah faidx instantiate records

AndreasHeger:


kyleabeauchamp:
LTGM.

Minor optional comment.  It might be possible to achieve the same thing without `set_x()` functions and instead via `@property` getters and setters.  The only downside there would be the inability to simultaneously set  sequence and qual, so that might still need a standalone member function with arguments.

AndreasHeger:
Thanks. My thoughts in not using @property this time was to not cause the same confusion as with AlignedSegment. There, the issue was that getter/setters were sometimes expensive operations and people got caught out. Caching every property is cumbersome. 
That being said, FastxRecord is light-weight and could provide @property access.

kyleabeauchamp:
Cool, good to know the details here.

PR #515, opened 2017-07-22
Refactor parse region

bioinformed:
## Scope

* Update glossary entry for CIGAR
* Correct error return in pysam_bam_update
* Remove incorrect use of inline declarations in external Cython definitions
* Refactor region_parsing and related code to HTSFile base class.

    Doing so exposed a bifurcation in terminology.  `AlignmentFile` uses
    `reference` and `end`, where as `VariantFile` is careful to always use
    `contig` and `stop`.  Although there is more code in the wild using
    `AlignmentFile`, I *strongly* prefer the latter nomenclature and updated all
    of the code to use that.  However, extra keyword arguments were added to
    enable backward compatible use of `reference` and `end`. 

I understand this change may be a bit controversial and I am happy to discuss before merging this PR.

### DIFF
--- a/doc/glossary.rst
+++ b/doc/glossary.rst
@@ -1,4 +1,4 @@
-========
+ ========
 Glossary
 ========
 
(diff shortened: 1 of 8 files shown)


AndreasHeger:
Thanks for the fixes. 
We should be consistent. I prefer contig over reference as well. Why 'stop' rather than 'end'? Neither are reserved words.

AndreasHeger:
... but generally, not an issue to go with contig, start, stop.

bioinformed:
It is a picky English thing. The most natural converse of 'end' is 'begin', not 'start'.  Also I've found that 'end' is more often linguistically correlated with inclusive coordinates than 'stop'. I've also standardized on 'start' and 'stop' in over 150k lines of code spanning many libraries and applications, so it is a strong stylistic convention that has taken root among several significant commercial NGS groups. 

I understand if these reasons are not as compelling to you as they are to me.

AndreasHeger:
Thanks, let us go with contig, start, stop .

bioinformed:
Great. I should mention that this PR is a precursor to adding support for retry logic around operations that occasionally fail over HTTP(S) and S3.  I'm going to focus initially on VariantFile and Fastafile, but will add the same logic to AlignmentFile when I have a stable solution to the former cases.

Issue #504, opened 2017-07-10
Data loss when fetching from S3/HTTP/HTTPS


bioinformed:
Using pysam 0.11.2.2 linked against libcurl (version tbd), I was fetching 4,000 BCF files via signed S3 HTTPS URLs one chromosome at a time by specifying both BCF and index URLs.  A very few number of those fetches failed, but either due to a failure in htslib or in pysam, the failure was not retried until success nor was the error propagated to the calling code.  This failure led to silent data loss.

I'm going to investigate more deeply and request clarification from the htslib developers on how error and retry handling are performed when htslib is linked against libcurl.

jmarshall:
HTSlib does no retrying, but propagates all errors to its callers (modulo bugs, but unlike some earlier HTSlib code, this code was written with correct error handling as an explicit goal).

(Were unreported transfer failures in BCF or index files? There is more older code involved with indices, so there is more scope there for less careful HTSlib code.)

bioinformed:
@jmarshall: Thanks for the timely reply.  I suspect the error may have been propagated via a NULL return from `hts_itr_querys`, which is currently treated as an empty query result in pysam.  I need to look more closely at why this was done, but I vaguely remember that queries for a valid contig with no resulting records could result in a NULL value from `hts_itr_querys`.

- See https://github.com/pysam-developers/pysam/blob/master/pysam/libcbcf.pyx#L3743

The error handling around fetching records from an iterator appears to be more sound.

- See https://github.com/pysam-developers/pysam/blob/master/pysam/libcbcf.pyx#L3788

While this failure appears to be due to pysam, it would be very nice if there could be more documentation on the iteration API.  It would also be ideal to have a method to determine which operations could be retried versus those that are doomed to always fail.

PR #516, opened 2017-07-23
Expose IO errors via IOError exceptions

bioinformed:
## Scope

Restructure error handling in `FastaFile.fetch`, `FastaFile._fetch`, and `VariantFile.fetch` to expose IO errors by raising Python`IOError` exceptions when `errno` is set.  The intent is to address hidden data loss (Issue #504), clarify when IO errors are occurring, and incremental to being able to add retry logic.  Testing is difficult, since simulating IO errors is not trivial.

Also added some missing detection of BCF allocation failures and raise `MemoryException`.

## Concerns

`FastaFile.fetch` still raises `ValueError` for non-IO errors.  However raising `IOError` is a non-backward compatible change and may cause problems in practice.  To me, is justifiable in order to unobfuscate when IO errors occur.

### DIFF
--- a/pysam/libcalignmentfile.pyx
+++ b/pysam/libcalignmentfile.pyx
@@ -1298,7 +1298,7 @@ cdef class AlignmentFile(HTSFile):
             if errno == EPIPE:
                 errno = 0
             else:
-                raise OSError(errno, force_str(strerror(errno)))
+                raise IOError(errno, force_str(strerror(errno)))
 
     def __dealloc__(self):
         cdef int ret = 0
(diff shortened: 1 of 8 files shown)


AndreasHeger:
Thanks, generally a good thing, but if not backwards compatible, shall we increase version to 0.13?

bioinformed:
## Scope

Restructure error handling in `FastaFile.fetch`, `FastaFile._fetch`, and `VariantFile.fetch` to expose IO errors by raising Python`IOError` exceptions when `errno` is set.  The intent is to address hidden data loss (Issue #504), clarify when IO errors are occurring, and incremental to being able to add retry logic.  Testing is difficult, since simulating IO errors is not trivial.

Also added some missing detection of BCF allocation failures and raise `MemoryException`.

## Concerns

`FastaFile.fetch` still raises `ValueError` for non-IO errors.  However raising `IOError` is a non-backward compatible change and may cause problems in practice.  To me, is justifiable in order to unobfuscate when IO errors occur.

Issue #517, opened 2017-07-23
Keep header when file is closed


bjoernpollex:
After closing an `AlignmenFile`, the `reference_name` attribute of `AlignedSegment` instances from that file cannot be accessed anymore, because the header data has been deleted. Example:

```python
from pysam import AlignmentFile

def load(path):
    with AlignmentFile(path) as bam:
        return next(bam)

al = load('some.bam')
al.reference_name
```

The last line results in the following error:

```
ValueError: I/O operation on closed file
```

Would it be possible to keep the header information around to support that operation after the file is closed? Would it be possible to store the reference names as Python strings in a tuple, and reference that in `AlignedSegment`? This way Python's own resource management mechanism would handle it.

bioinformed:
@AndreasHeger: Should we consider splitting the `AlignmentHeader` from `AlignmentFile` akin to what is done for `VariantFile` and `VariantHeader`?  The overhead of storing a reference to the header is fairly minor and can finally remove the need for having to even think about translating to and from `tid`s.

AndreasHeger:
@bjoernpollex thanks for reporting and @bioinformed, yes. I realize not providing a full abstraction on top of htslib requires a lot of understandable confusion, such as here #509.
I will do this today.

bjoernpollex:
Thanks, this is really an essential project to our work, and the support is much appreciated!

PR #518, opened 2017-07-24
{AH} separate header into AlignmentHeader from AlignmentFile, closes …

AndreasHeger:
…#517

This PR separates header functionality into a separate class (AlignmentHeader, cf. VariantHeader) and propagates the information through to iterators and AlignedSegments. The benefit is that:

1. reference names can be looked up even after a file has been closed or is out-of-scope
2. reference names are always available and no need to look up via tid.

At the moment AlignmentHeaders are read-only objects, i.e. the header needs to be created before any derived AlignedSegments are being created. 

A few backwards incompatible changes are the consequence of this refactoring, most notable that AlignmentFile.header now returns AlignmentHeader and not a dictionary.

### DIFF
--- a/pysam/libcalignedsegment.pxd
+++ b/pysam/libcalignedsegment.pxd
@@ -79,7 +79,7 @@ cdef class PileupRead:
     cdef uint32_t _is_refskip
 
 # factory methods
-cdef makeAlignedSegment(bam1_t * src, AlignmentHeader header)
-cdef makePileupColumn(bam_pileup1_t ** plp, int tid, int pos, int n_pu, AlignmentHeader header)
-cdef makePileupRead(bam_pileup1_t * src, AlignmentHeader header)
+cdef AlignedSegment makeAlignedSegment(bam1_t * src, AlignmentHeader header)
+cdef PileupColumn makePileupColumn(bam_pileup1_t ** plp, int tid, int pos, int n_pu, AlignmentHeader header)
+cdef PileupRead makePileupRead(bam_pileup1_t * src, AlignmentHeader header)
 cdef uint32_t get_alignment_length(bam1_t * src)
--- a/pysam/libcalignedsegment.pxd
+++ b/pysam/libcalignedsegment.pxd
@@ -37,7 +37,7 @@ cdef class AlignedSegment:
     cdef bam1_t * _delegate
 
     # the header that a read is associated with
-    cdef AlignmentHeader header
+    cdef readonly AlignmentHeader header
 
     # caching of array properties for quick access
     cdef object cache_query_qualities
--- a/pysam/libcalignedsegment.pxd
+++ b/pysam/libcalignedsegment.pxd
@@ -26,7 +26,7 @@ cdef extern from "htslib_util.h":
     void pysam_update_flag(bam1_t * b, uint16_t v, uint16_t flag)
 
 
-from pysam.libcalignmentfile cimport AlignmentFile
+from pysam.libcalignmentfile cimport AlignmentFile, AlignmentHeader
 ctypedef AlignmentFile AlignmentFile_t
 
 
@@ -36,8 +36,8 @@ cdef class AlignedSegment:
     # object that this AlignedSegment represents
     cdef bam1_t * _delegate
 
-    # the file from which this AlignedSegment originates (can be None)
-    cdef AlignmentFile _alignment_file
+    # the header that a read is associated with
+    cdef AlignmentHeader header
 
     # caching of array properties for quick access
     cdef object cache_query_qualities
@@ -57,7 +57,7 @@ cdef class AlignedSegment:
     cpdef has_tag(self, tag)
 
     # returns a valid sam alignment string
-    cpdef tostring(self, AlignmentFile_t handle)
+    cpdef tostring(self, htsfile=*)
 
 
 cdef class PileupColumn:
@@ -65,7 +65,7 @@ cdef class PileupColumn:
     cdef int tid
     cdef int pos
     cdef int n_pu
-    cdef AlignmentFile _alignment_file
+    cdef AlignmentHeader header
 
 
 cdef class PileupRead:
@@ -78,8 +78,8 @@ cdef class PileupRead:
     cdef uint32_t _is_tail
     cdef uint32_t _is_refskip
 
-# factor methods
-cdef makeAlignedSegment(bam1_t * src, AlignmentFile alignment_file)
-cdef makePileupColumn(bam_pileup1_t ** plp, int tid, int pos, int n_pu, AlignmentFile alignment_file)
-cdef makePileupRead(bam_pileup1_t * src, AlignmentFile alignment_file)
+# factory methods
+cdef makeAlignedSegment(bam1_t * src, AlignmentHeader header)
+cdef makePileupColumn(bam_pileup1_t ** plp, int tid, int pos, int n_pu, AlignmentHeader header)
+cdef makePileupRead(bam_pileup1_t * src, AlignmentHeader header)
 cdef uint32_t get_alignment_length(bam1_t * src)
--- a/pysam/libcalignedsegment.pyx
+++ b/pysam/libcalignedsegment.pyx
@@ -486,27 +486,27 @@ cdef inline object getQualitiesInRange(bam1_t *src,
 
 
 #####################################################################
-## private factory methods
+## factory methods for instantiating extension classes
 cdef class AlignedSegment
-cdef makeAlignedSegment(bam1_t * src, AlignmentFile alignment_file):
+cdef makeAlignedSegment(bam1_t * src, AlignmentHeader header):
     '''return an AlignedSegment object constructed from `src`'''
     # note that the following does not call __init__
     cdef AlignedSegment dest = AlignedSegment.__new__(AlignedSegment)
     dest._delegate = bam_dup1(src)
-    dest._alignment_file = alignment_file
+    dest.header = header
     return dest
 
 
 cdef class PileupColumn
 cdef makePileupColumn(bam_pileup1_t ** plp, int tid, int pos,
-                      int n_pu, AlignmentFile alignment_file):
+                      int n_pu, AlignmentHeader header):
     '''return a PileupColumn object constructed from pileup in `plp` and
     setting additional attributes.
 
     '''
     # note that the following does not call __init__
     cdef PileupColumn dest = PileupColumn.__new__(PileupColumn)
-    dest._alignment_file = alignment_file
+    dest.header = header
     dest.plp = plp
     dest.tid = tid
     dest.pos = pos
@@ -514,10 +514,10 @@ cdef makePileupColumn(bam_pileup1_t ** plp, int tid, int pos,
     return dest
 
 cdef class PileupRead
-cdef inline makePileupRead(bam_pileup1_t * src, AlignmentFile alignment_file):
+cdef inline makePileupRead(bam_pileup1_t * src, AlignmentHeader header):
     '''return a PileupRead object construted from a bam_pileup1_t * object.'''
     cdef PileupRead dest = PileupRead.__new__(PileupRead)
-    dest._alignment = makeAlignedSegment(src.b, alignment_file)
+    dest._alignment = makeAlignedSegment(src.b, header)
     dest._qpos = src.qpos
     dest._indel = src.indel
     dest._level = src.level
@@ -757,10 +757,10 @@ cdef class AlignedSegment:
                                    self.tags)))
 
     def __copy__(self):
-        return makeAlignedSegment(self._delegate, self._alignment_file)
+        return makeAlignedSegment(self._delegate, self.header)
 
     def __deepcopy__(self, memo):
-        return makeAlignedSegment(self._delegate, self._alignment_file)
+        return makeAlignedSegment(self._delegate, self.header)
 
     def compare(self, AlignedSegment other):
         '''return -1,0,1, if contents in this are binary
@@ -824,7 +824,7 @@ cdef class AlignedSegment:
 
         return hash_value
 
-    cpdef tostring(self, AlignmentFile_t htsfile):
+    cpdef tostring(self, htsfile=None):
         """returns a string representation of the aligned segment.
 
         The output format is valid SAM format.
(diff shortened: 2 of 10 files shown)


pysam/libcalignedsegment.pyx
```diff
@@ -486,38 +486,38 @@ cdef inline object getQualitiesInRange(bam1_t *src,
 
 
 #####################################################################
-## private factory methods
+## factory methods for instantiating extension classes
 cdef class AlignedSegment
-cdef makeAlignedSegment(bam1_t * src, AlignmentFile alignment_file):
+cdef makeAlignedSegment(bam1_t * src, AlignmentHeader header):
```
bioinformed: I prefer the convention where there is no space after `*`.  It seems the most common convention in well written C code and what I've been emulating in the BCF code.

pysam/libcalignedsegment.pyx
```diff
@@ -934,15 +927,16 @@ cdef class AlignedSegment:
...
             the sequence dictionary. To obtain the name of the
             reference sequence, use
-            :meth:`pysam.AlignmentFile.getrname()`
+            :meth:`get_reference_name()`
 
         """
         def __get__(self): return self._delegate.core.tid
         def __set__(self, tid): self._delegate.core.tid = tid
 
     property reference_start:
```
bioinformed: Should convert to Python property syntax at some point.  Just raising the point-- does not need to be done.

pysam/libcalignedsegment.pyx
```diff
@@ -2448,8 +2442,9 @@ cdef class PileupColumn:
             # warning: there could be problems if self.n and self.buf are
             # out of sync.
             for x from 0 <= x < self.n_pu:
-                pileups.append(makePileupRead(&(self.plp[0][x]),
-                                              self._alignment_file))
+                pileups.append(
```
bioinformed: I'm having trouble justifying why 79/80 column limits are still relevant in this day and age.  My laptop display can show 150+ columns with my default font.  Can we decide to defy PEP8 a little and support a 120 column limit in pysam?

pysam/libcalignedsegment.pyx
```diff
@@ -2448,8 +2442,9 @@ cdef class PileupColumn:
...
+                    makePileupRead(&(self.plp[0][x]),
```
bioinformed: p.s. Pretty sure there is no need for parens around `&self.plp[0][x]`, since `&` has lower precedence than `[]`.

pysam/libcalignmentfile.pxd
```diff
@@ -77,7 +80,7 @@ cdef class IteratorRow:
     cdef bam1_t * b
     cdef AlignmentFile samfile
     cdef htsFile * htsfile
-    cdef bam_hdr_t * header
+    cdef readonly AlignmentHeader header
```
bioinformed: Does this need to be exposed at all?

pysam/libcalignmentfile.pxd
```diff
@@ -120,7 +125,7 @@ cdef class IteratorColumn:
     cdef bam_pileup1_t * plp
     cdef bam_plp_t pileup_iter
     cdef __iterdata iterdata
-    cdef AlignmentFile samfile
+    cdef readonly AlignmentFile samfile
```
bioinformed: ditto

pysam/libcalignmentfile.pyx
```diff
@@ -213,20 +241,26 @@ cdef bam_hdr_t * build_header_from_dict(new_header):
...
+# are not implemented in cython.
+cdef AlignmentHeader makeAlignmentHeader_from_list(reference_names,
+                                                   reference_lengths,
+                                                   add_sq_text=True,
+                                                   text=None):
+    """build header from list of reference names and lengths.
+    """
+    # create new header
+    cdef AlignmentHeader new_header = AlignmentHeader.__new__(AlignmentHeader)
+    new_header.ptr = bam_hdr_init()
```
bioinformed: Should check `ptr` and raise `MemoryError` if NULL.

pysam/libcalignmentfile.pyx
```diff
@@ -550,16 +771,16 @@ cdef class AlignmentFile(HTSFile):
 
             # header structure (used for writing)
             if template:
-                self.header = bam_hdr_dup(template.header)
+                self.header = template.header.copy()
```
bioinformed: Should check type of `template.header`.  Also, `copy` is not needed at the moment, since `AlignmentHeader`s are immutable.  May still be a good idea to keep the `copy` if/when they do change.

pysam/libcalignmentfile.pyx
```diff
@@ -1348,7 +1541,7 @@ cdef class AlignmentFile(HTSFile):
```
bioinformed: Add tid translation here for when `read.header is not self.header`?  And add a `translate` method to `AlignedSegment`.

This is a breaking change that needs to happen at some point, because it breaks apps that have already translated their tids.

pysam/libcalignedsegment.pyx
```diff
@@ -486,38 +486,38 @@ cdef inline object getQualitiesInRange(bam1_t *src,
...
+## factory methods for instantiating extension classes
 cdef class AlignedSegment
-cdef makeAlignedSegment(bam1_t * src, AlignmentFile alignment_file):
+cdef makeAlignedSegment(bam1_t * src, AlignmentHeader header):
     '''return an AlignedSegment object constructed from `src`'''
     # note that the following does not call __init__
     cdef AlignedSegment dest = AlignedSegment.__new__(AlignedSegment)
     dest._delegate = bam_dup1(src)
-    dest._alignment_file = alignment_file
+    dest.header = header
```
bioinformed: must raise an exception if `header` is false (None or destroyed).

pysam/libcalignmentfile.pyx
```diff
@@ -273,7 +307,193 @@ cdef bam_hdr_t * build_header_from_list(reference_names,
...
+            bam_hdr_destroy(self.ptr)
+            self.ptr = NULL
+
+    def __bool__(self):
+        return self.ptr != NULL
+
+    def copy(self):
+        return makeAlignmentHeader(bam_hdr_dup(self.ptr))
+
+    def clear(self):
```
bioinformed: Strongly recommend not allowing a header to be cleared unless deallocated.  Allowing `ptr = NULL` in a live header will do more than invalidate objects, it will cause segfaults when accessing those objects.

pysam/libcalignmentfile.pyx
```diff
@@ -273,7 +307,193 @@ cdef bam_hdr_t * build_header_from_list(reference_names,
...
+        if self.ptr:
+            bam_hdr_destroy(self.ptr)
+            self.ptr = NULL
+    
+    property nreferences:
+        """"int with the number of :term:`reference` sequences in the file.
+
+        This is a read-only attribute."""
+        def __get__(self):
+            if self.ptr:
```
bioinformed: No need to check `self.ptr` if live `AlignmentHeader` objects require it.

pysam/libcalignmentfile.pyx
```diff
@@ -273,7 +307,193 @@ cdef bam_hdr_t * build_header_from_list(reference_names,
...
+            else:
+                return 0
+
+    property references:
+        """tuple with the names of :term:`reference` sequences. This is a
+        read-only attribute"""
+        def __get__(self):
+            t = []
+            if self.ptr:
+                for x from 0 <= x < self.ptr.n_targets:
```
bioinformed: probably also worth updating to `range()` syntax, rather than the archaic pyrex syntax.

pysam/libcalignmentfile.pyx
AndreasHeger: ok

pysam/libcalignmentfile.pyx
AndreasHeger: ok

pysam/libcalignmentfile.pxd
AndreasHeger: Thanks, will go through all of these.

pysam/libcalignedsegment.pyx
AndreasHeger: ... just habit, 120 is ok. Which reminds me that I want to turn on pep8 tests at one point.

I am short-sighted and usually work with a extra large font-size and an emacs split screen having two windows side-by-side (such as code | tests). Need better glasses :=).

bioinformed: I should probably disclose that my desktop monitor is 34" wide and can show 3 screens with >140 columns side-by-side.  I realize that is a bit extreme, but wonderful for running IGV.

pysam/libcalignedsegment.pxd
```diff
@@ -57,15 +57,15 @@ cdef class AlignedSegment:
     cpdef has_tag(self, tag)
 
     # returns a valid sam alignment string
-    cpdef tostring(self, AlignmentFile_t handle)
+    cpdef tostring(self, htsfile=*)
```
bioinformed: Not clear why this method needs to be `cpdef`.  Do we call it often from a `cdef` function?

pysam/libcalignmentfile.pyx
AndreasHeger: done

pysam/libcalignmentfile.pyx
AndreasHeger: Removed clear()

pysam/libcalignmentfile.pyx
AndreasHeger: done

pysam/libcalignmentfile.pyx
AndreasHeger: done

pysam/libcalignmentfile.pyx
AndreasHeger: Ok. Leaving copy in for the moment, but commented the code.

pysam/libcalignmentfile.pyx
```diff
@@ -550,16 +771,16 @@ cdef class AlignmentFile(HTSFile):
 
             # header structure (used for writing)
             if template:
-                self.header = bam_hdr_dup(template.header)
+                self.header = template.header.copy()
             elif header:
-                self.header = build_header_from_dict(header)
+                self.header = makeAlignmentHeader_from_dict(header)
```
AndreasHeger: done

pysam/libcalignmentfile.pyx
```diff
@@ -1579,21 +1631,82 @@ cdef class AlignmentFile(HTSFile):
...
+            return makeAlignedSegment(self.b, self.header)
         elif ret == -2:
             raise IOError('truncated file')
         else:
             raise StopIteration
 
+    ###########################################################################
+    # methods/properties referencing the header. These are mostly for backwards
+    # compatibility for pysam < 0.12
+    def is_valid_tid(self, int tid):
```
AndreasHeger: ok, removed misleading comment

pysam/libcalignmentfile.pyx
```diff
@@ -2418,34 +2540,39 @@ cdef class IndexedReads:
...
                 self.htsfile = hts_open(cfilename, 'r')
             assert self.htsfile != NULL
-            # read header - required for accurate positioning
+
+            # need to advance in newly opened file to position after header
+            # better: use seek/tell?
             with nogil:
-                self.header = sam_hdr_read(self.htsfile)
+                hdr = sam_hdr_read(self.htsfile)
+            assert hdr != NULL
```
AndreasHeger: done

tests/AlignmentFile_test.py
```diff
@@ -733,13 +734,36 @@ def testClosedFile(self):
...
-
+    def test_header_available_after_closing_file(self):
+
+        def load_bam():
+            with pysam.AlignmentFile(os.path.join(BAM_DATADIR, "ex1.bam"), "rb") as inf:
+                header = inf.header
+            return header
+
+        header = load_bam()
+        self.assertTrue(header != None)
```
AndreasHeger: done

pysam/libcalignedsegment.pyx
AndreasHeger: ok

pysam/libcalignedsegment.pyx
AndreasHeger: done

pysam/libcalignedsegment.pyx
AndreasHeger: might misunderstand, but reference_name property does already exist in AlignedSegment

pysam/libcalignedsegment.pyx
AndreasHeger: done

pysam/libcalignmentfile.pxd
AndreasHeger: removed

pysam/libcalignmentfile.pxd
AndreasHeger: removed

pysam/libcalignedsegment.pyx
bioinformed: My suggestion was to add validating *setters* for `reference_name` and `reference_id`.  `reference_name` has no setter and `reference_id` has a non-validating setter.

AndreasHeger: Ah, thanks.

AndreasHeger: I will add these as now is the right time.

pysam/libcalignmentfile.pyx
```diff
@@ -1766,7 +1761,8 @@ cdef class IteratorRow:
             # better: use seek/tell?
             with nogil:
                 hdr = sam_hdr_read(self.htsfile)
-            assert hdr != NULL
+            if hdr is NULL:
```
bioinformed: I'm not sure how Cython treats tests for object identity with `NULL`.  I typically use a boolean test instead: `if not hdr`.

pysam/libcalignmentfile.pyx
```diff
@@ -273,7 +301,230 @@ cdef bam_hdr_t * build_header_from_list(reference_names,
...
+    # Python constructor
+    def __init__(self,
+                 reference_names=None,
+                 reference_lengths=None,
+                 text=None,
+                 header_dict=None,
+                 add_sq_text=True):
+
+        self.ptr = bam_hdr_init()
+        if self.ptr is NULL:
```
bioinformed: `is` -> `==`

AndreasHeger:
TODO, @bioinformed suggestion:

... add an option to the AlignmentFile constructor which allows or restricts passing the AlignmentHeader to AlignedSegments.  If the header is None, then tids are handled as they are now.  If not, there will be intelligent setting and remapping as aligned segments are passed between AlignmentFiles

Issue #443, opened 2017-04-19
0.11.0 build failure with missing symbols


nieder:
Building 0.11.0 on OS X with htslib1.4 leads to missing symbols when linking libcsamtools.so:
~~~~
[...]

PR #519, opened 2017-07-28
fix missing symbols in libcsamtools and libcbcftools libraries

vorlonofportland:
Building on Ubuntu fails because a couple of the .so's wind up with
unresolved references to symbols found elsewhere in the tree.  Fix their
linkage so that they load cleanly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -463,6 +463,7 @@ cutils = Extension(
 csamtools = Extension(
     "pysam.libcsamtools",
     [source_pattern % "samtools"] +
+    ["pysam/pysam_util.c"] +
     glob.glob(os.path.join("samtools", "*.pysam.c")) +
     htslib_sources +
     os_c_files,
@@ -484,7 +485,8 @@ cbcftools = Extension(
     library_dirs=["pysam"] + htslib_library_dirs,
     include_dirs=["bcftools", "pysam", "."] +
     include_os + htslib_include_dirs,
-    libraries=external_htslib_libraries + internal_htslib_libraries,
+    libraries=external_htslib_libraries + internal_htslib_libraries + \
+    [os.path.splitext("csamtools{}".format(suffix))[0]],
     language="c",
     extra_compile_args=extra_compile_args,
     define_macros=define_macros


bioinformed:
@vorlonofportland: What version of Ubuntu are you seeing this on?

vorlonofportland:
@bioinformed this is for the upcoming Ubuntu 17.10 release, building against artful-proposed (new htslib is currently blocked from the Ubuntu release due to pysam compatibility).

I haven't checked whether this is specific to Ubuntu, or if it's also reproducible in Debian unstable. I believe it would be, since I don't think Ubuntu has any non-default linker flags that change this behavior.

bioinformed:
@vorlonofportland: Thanks for the information.  Do you have any idea why linking is now different under Ubuntu 17.10?  This doesn't seem to be a problem on previous versions.

I'll test this branch on OS X, CentOS, and BSD to be sure there are no other side effects.

bioinformed:
and look into why `testAllStepper` is failing.

AndreasHeger:
Thanks @vorlonofportland and @bioinformed

bioinformed:
I've been able to reproduce the crash on OS X:
```
error: libcsamtools.cpython-36m-darwin.so debug map object file '/Users/kevin.jacobs/projects/pysam/build/temp.macosx-10.12-x86_64-3.6/pysam/pysam_util.o' has changed (actual time is 0x59849f51, debug map time is 0x59849f34) since this executable was linked, file will be ignored
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x68)
  * frame #0: 0x00007fffdbd7f6a0 libsystem_c.dylib`flockfile + 4
    frame #1: 0x00007fffdbd88324 libsystem_c.dylib`vfprintf_l + 28
    frame #2: 0x00007fffdbd80dff libsystem_c.dylib`fprintf + 186
    frame #3: 0x0000000103c9d2df libcsamtools.cpython-36m-darwin.so`mpileup(conf=<unavailable>, n=1, fn=0x0000000101c18980) at bam_plcmd.c.pysam.c:418 [opt]
    frame #4: 0x0000000103c9ca1b libcsamtools.cpython-36m-darwin.so`bam_mpileup(argc=<unavailable>, argv=<unavailable>) at bam_plcmd.c.pysam.c:1114 [opt]
    frame #5: 0x0000000103c757eb libcutils.cpython-36m-darwin.so`__pyx_pf_5pysam_9libcutils_8_pysam_dispatch(__pyx_self=<unavailable>, __pyx_v_collection=<unavailable>, __pyx_v_method=0x0000000000000007, __pyx_v_args=<unavailable>, __pyx_v_catch_stdout=<unavailable>, __pyx_v_is_usage=<unavailable>, __pyx_v_save_stdout=<unavailable>) at libcutils.c:6876 [opt]
    frame #6: 0x0000000100054420 python`_PyCFunction_FastCallDict + 464
[...]
```

But have no specific insight as to why it is currently failing.  Any thoughts, @vorlonofportland ?

AndreasHeger:
Revisiting this PR - I think the fundamental problem is that the dependencies have become too complex. I will try to simplify.

AndreasHeger:
Thanks, the issue has been variables declared as extern between cython extensions. Please see comment to #443.
I am closing this PR now, but please let me know if there are unresolved issues.

PR #520, opened 2017-07-28
skip usage() tests on autobuilders

vorlonofportland:
The usage() functions check whether stdin is a tty before displaying usage
information.  Therefore, we should not test this code when running on an
autobuilder without a tty.

Clever allocation of a pty from within the test suite itself for purposes
of testing this path is left as an exercise for the reader.


AndreasHeger:
Great, thanks

PR #521, opened 2017-07-31
Simplify file detection logic and allow remote index files

bioinformed:
## Scope

Substantially simplify index loading logic in `AlignmentFile._open`:

* Remove attempts to guess data and index file names; this is magic that I recommend we leave to htslib.
* Remove file existence check prior to opening files with htslib; do check for errors more carefully afterward and raise the appropriate error (`IOError` for when `errno` is set, `ValueError` otherwise for backward compatibility)
* Catch IO errors when explicitly loading an index.
* Allow remote indices (tested using S3 signed URLs, not easily added to the test suite).
* Document `filepath_index` and make it an alias for `index_filename`.
* Added a `require_index` parameter to `AlignmentFile`

### DIFF
--- a/doc/release.rst
+++ b/doc/release.rst
@@ -7,6 +7,17 @@ Release 0.12.0
 
 * [#473] A new FastxRecord class that can be instantiated from class and
   modified in-place. Replaces PersistentFastqProxy.
+* [#521] In AligmentFile, Simplify file detection logic and allow remote index files
+  * Removed attempts to guess data and index file names; this is magic left
+    to htslib.
+  * Removed file existence check prior to opening files with htslib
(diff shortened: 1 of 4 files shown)


AndreasHeger:
Great, thanks. Checking for indices is historical baggage, the more we can remove, the better!

cshenanigans:
@bioinformed 👍 awesome, I need this! any idea when you'll be releasing 0.12? P.S. Hi!!!!!

bioinformed:
@cshenanigans: Assuming we can resolve the last few bugs, the plan is to release in the next day or two.  Please let us know what other features you'd like to see.

PR #526, opened 2017-08-08
Handle unset ref when creating new records

msto:
Hi,

I found a small bug in the updated `VariantHeader.new_record` method. When setting the record's initial attributes, the setters' calls to `bcf_sync_end` raise an exception because `record.ref` is `None`. 

I added a quick check in `bcf_sync_end` to handle this case. I think setting `ref_len` to 0 prior to the INFO/END deletion shouldn't have any side effects, since a new record shouldn't have any alleles or a set END anyways. 

This bug could also be solved by setting `ref` before setting the remaining attributes in `new_record`, but I wasn't sure what the desired default would be for `ref` (or if it would be a good idea to force a default).

### DIFF
--- a/pysam/libcbcf.pyx
+++ b/pysam/libcbcf.pyx
@@ -1159,7 +1159,13 @@ cdef inline bcf_sync_end(VariantRecord record):
     cdef bcf_hdr_t *hdr = record.header.ptr
     cdef bcf_info_t *info
     cdef int end_id = bcf_header_get_info_id(record.header.ptr, b'END')
-    cdef int ref_len = len(record.ref)
+    cdef int ref_len 
+
+    # allow missing ref when instantiating a new record
+    if record.ref is not None:
+        ref_len = len(record.ref)
+    else:
+        ref_len = 0
 
     # Delete INFO/END if no alleles are present or if rlen is equal to len(ref)
     if not record.ptr.n_allele or record.ptr.rlen == ref_len:


bioinformed:
Thanks for the bug report and PR!  Much appreciated!

msto:
No problem, thanks for merging it in so quickly!

I was actually going to follow up with a related bug. In the `ref` setter method, if the record doesn't already have an associated list of alleles, its list of alleles is set to contain just the ref allele.

https://github.com/pysam-developers/pysam/blob/27d84526c13a8ffb8231b5c2db713bc7341ffdd5/pysam/libcbcf.pyx#L3136-L3141

However, the `alleles` setter method raises an exception if passed a list of fewer than 2 alleles.

https://github.com/pysam-developers/pysam/blob/27d84526c13a8ffb8231b5c2db713bc7341ffdd5/pysam/libcbcf.pyx#L3169-L3170 

How do you think this one should be solved? Is the allele count restriction necessary? I ran a quick test without it and nothing seemed to break, so I could submit another PR with a quick fix removing that test. If it is necessary, I think setting a new record's ref allele would then require the addition of a dummy or default alternate allele.

bioinformed:
VCF doesn't allow null alleles and requires at least one alt allele.  The easier way to proceed is to require `ref` and at least one `alt` to be specified during construction (even if they default to `'N'` and `'<NON_REF>'`) and then enforce that requirement globally.

PR #532, opened 2017-09-03
Allow for building out of tree

SoapGentoo:
* Gentoo builds all python packages out of tree,
  as is best practice. Using `-Isamtools` implicitly
  assumes that everything is being built inplace.

  See also: https://bugs.gentoo.org/show_bug.cgi?id=629670

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -359,6 +359,8 @@ else:
 
 define_macros = []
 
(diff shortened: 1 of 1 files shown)


AndreasHeger:
Great, many thanks!

PR #533, opened 2017-09-03
Create bcftools/config.h if it doesn't exist

0xaf1f:
since bcftools and samtools 1.5 are using the same build system,
they need to be treated the same way. Compilation fails when bcftools
can't find its config.h.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -10,6 +10,7 @@ include KNOWN_BUGS
 include THANKS
 include cy_build.py
 include requirements.txt
+exclude bcftools/config.h
 include pysam/libc*.pxd
 include pysam/libc*.pyx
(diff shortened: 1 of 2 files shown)


AndreasHeger:
@0xaf1f - I am really sorry I did not notice this pull-request - it would have saved me quite some time.
I was away and swamped by github notifications and wanted to make sure the conda package builds as soon as possible, but did not check the pull requests. 
Many thanks for submitting this - lesson learned!

AndreasHeger:
I am closing the PR as it indeed exactly does the same fixes, see  a67b90f

0xaf1f:
No worries. I guess if I checked the issues more closely, I could have made it known there that I had found the solution.

PR #537, opened 2017-09-12
Allow tabix index files to be created in a custom location

mvdbeek:
This is one of the modifications we have been using in [Galaxy](https://github.com/galaxyproject/galaxy), where we can't easily store indexes next to the data files they belong to.

With the recent pysam changes that allow custom index locations we thought it may be of interest to also allow creating indexes in non-default locations.

This PR also includes an option that prevents deleting uncompressed tabix files using the `keep_original` argument when generating an index (we consider a users history as immutable, so we would like to avoid removing files), and we check if a file is compressed using gzip instead of assuming that gzipped files must end with `.gz` (since our storage does not include meaningful filenames).


pysam/libctabix.pyx
```diff
@@ -844,6 +844,7 @@ def tabix_compress(filename_in,
 
 
 def tabix_index(filename, 
+                index=None,
```
nsoranzo: It's probably better to add this as last parameter.

AndreasHeger:
Great, many thanks!

bgruening:
:+1:

PR #561, opened 2017-10-30
Assert on reads with invalid MD tags

ramyala:
Its possible that MD tag derived seqeunce length is longer than CIGAR
alignment length. In such cases, building alignment sequence can lead
to a buffer overflow error. Such reads will now generate an assertion.


AndreasHeger:
Great, thanks!

PR #565, opened 2017-11-07
Improve assert log for mismatched MD tag reads

ramyala:
Include insertion length along with the MD and CIGAR details in assert error


AndreasHeger:
Thanks

PR #566, opened 2017-11-07
Retry logic on htslib seek failure

ramyala:
This is useful when htslib seek fails due to network resource issues.

The retry logic is hid behind pysam so as to make fetch() calls seemless.

### DIFF
--- a/pysam/libcalignmentfile.pxd
+++ b/pysam/libcalignmentfile.pxd
@@ -53,6 +53,10 @@ cdef class AlignmentFile(HTSFile):
     # write an aligned read
     cpdef int write(self, AlignedSegment read) except -1
 
+    # retry settings on network read failure
+    cdef uint32_t _max_retries
+    cdef uint32_t _wait_time_base
+    cdef uint32_t _wait_multiplier
 
 cdef class PileupColumn:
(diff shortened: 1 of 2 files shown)


bioinformed:
Nice!  I'll do some testing and see if I can generalize the logic in the `HTSFile` base class, since other `HTSFile` types can benefit from the same functionality.

AndreasHeger:
Thanks both. @bioinformed, can you merge when done? Thanks

daviesrob:
Note that this depends on some changes in samtools/htslib#606, which is unlikely to be merged as it is.

As far as I can make out, the problem here is that seeking on HTTP(S) streams involves opening a new connection (and dropping the old one) on each seek.  This means you can get rate limited if you try to seek to a lot of locations very quickly.

We have some ongoing work to mitigate this in htslib by improving BGZF block caching, and converting small forward seeks to reads.  It might also be possible to implement some form of retry in `hfile_libcurl`.

If the above scenario is correct, then I suspect the quickest solution would be download the (parts of) files you want to local storage, and then work on those copies.  It's likely to be much kinder on the upstream server.

ramyala:
I have patches to optimize libcurl seek. I'll send out PRs.

PR #570, opened 2017-11-16
Provide default alternate allele when constructing a new VariantRecord

msto:
Hi,

I found a bug where it wasn't possible to set the `ref` of a newly constructed VariantRecord. In the absence of any defined alleles, the ref setter method sets the record's alleles to be the list of only the reference allele. However, the alleles setter method requires at least two alleles to be passed. I fixed this by providing a generic "NON_REF" allele along with the ref allele if no others are present.

### DIFF
--- a/pysam/libcbcf.pyx
+++ b/pysam/libcbcf.pyx
@@ -1185,13 +1185,15 @@ cdef inline bcf_sync_end(VariantRecord record):
         bcf_info_set_value(record, b'END', record.ptr.pos + record.ptr.rlen)
 
 
-cdef inline bint has_symbolic_allele(VariantRecord record):
-    # Check if alternate allele is symbolic
-    if record.ptr.n_allele >= 2:
-        alt = record.ptr.d.allele[1]
-        return (alt[0] == b'<' and alt[len(alt) - 1] == b'>')
-    else:
-        return False
+cdef inline int has_symbolic_allele(VariantRecord record):
+    """Return index of first symbolic allele. 0 if no symbolic alleles."""
+
+    for i in range(1, record.ptr.n_allele):
+        alt = record.ptr.d.allele[i]
+        if alt[0] == b'<' and alt[len(alt) - 1] == b'>':
+            return i
+
+    return 0
 
 
 ########################################################################
--- a/pysam/libcbcf.pyx
+++ b/pysam/libcbcf.pyx
@@ -1167,8 +1167,16 @@ cdef inline bcf_sync_end(VariantRecord record):
     else:
         ref_len = 0
 
+    # Check if alternate allele is symbolic
+    if record.ptr.n_allele >= 2:
+        alt = record.ptr.d.allele[1]
+        is_symbolic = (alt[0] == b'<' and alt[len(alt) - 1] == b'>')
+    else:
+        is_symbolic = False
+
     # Delete INFO/END if no alleles are present or if rlen is equal to len(ref)
-    if not record.ptr.n_allele or record.ptr.rlen == ref_len:
+    # Always keep END for symbolic alleles
+    if not is_symbolic and (not record.ptr.n_allele or record.ptr.rlen == ref_len):
         # If INFO/END is not defined in the header, it doesn't exist in the record
         if end_id >= 0:
             info = bcf_get_info(hdr, record.ptr, b'END')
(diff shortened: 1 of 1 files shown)


msto:
Hi,

I also added a patch to fix the behavior mentioned in #571 . In this patch, a VariantRecord's `rlen` will only be updated upon setting ref/alt alleles if its alternate allele isn't symbolic (angle-bracketed). Additionally, I updated `bcf_sync_end` so it won't remove `END` if the alternate allele is symbolic.

I thought about adding an `is_symbolic` or `has_symbolic_allele` property to the VariantRecord, but I wasn't sure if it was worth adding. If you'd like me to move that logic to such a property, let me know and I can update.

Matt

bioinformed:
Thanks, Matt!  This is a super useful change.  Can you add an inline Cython function to detect symbolic alleles?  I'd feel better with that logic centralized.  Otherwise, I'll do some testing and make sure there are no unforeseen edge cases.

msto:
Happy to help! I did find a small bug. The call to `bcf_update_alleles_str` on line 3196 changes `rlen` to the value of `END`. I'm not sure why it's doing that, but the easiest fix was to just cache the rlen for records with symbolic alleles. I fixed that and moved the symbolic allele check to an inline function per your request.

pysam/libcbcf.pyx
```diff
@@ -1184,6 +1185,15 @@ cdef inline bcf_sync_end(VariantRecord record):
         bcf_info_set_value(record, b'END', record.ptr.pos + record.ptr.rlen)
 
 
+cdef inline bint has_symbolic_allele(VariantRecord record):
+    # Check if alternate allele is symbolic
+    if record.ptr.n_allele >= 2:
```
bioinformed: The code below has to be a loop, since any of the alt alleles could be symbolic.

pysam/libcbcf.pyx
```diff
@@ -1184,6 +1185,15 @@ cdef inline bcf_sync_end(VariantRecord record):
         bcf_info_set_value(record, b'END', record.ptr.pos + record.ptr.rlen)
 
 
+cdef inline bint has_symbolic_allele(VariantRecord record):
```
bioinformed: change return type to be `int` and return the index of the first symbolic allele or zero if none exist.  This preserves the boolean nature of the function (0==no symbolic alleles, !=0 is the index if the first) and provides a more informative return type.

msto:
Hey Kevin,

Thanks for the helpful suggestions, I updated the code accordingly. Let me know if there are any other changes you'd like to see.

Matt

bioinformed:
Thanks!  Will run more tests and let you know.

pysam/libcbcf.pyx
```diff
@@ -1184,6 +1185,17 @@ cdef inline bcf_sync_end(VariantRecord record):
         bcf_info_set_value(record, b'END', record.ptr.pos + record.ptr.rlen)
 
 
+cdef inline int has_symbolic_allele(VariantRecord record):
+    """Return index of first symbolic allele. 0 if no symbolic alleles."""
+
+    for i in range(1, record.ptr.n_allele):
+        alt = record.ptr.d.allele[i]
```
bioinformed: Does `alt` need a cdef?  (This requires checking the generated Cython)

msto: Sorry, we're hitting the limits of my Cython here. How do you tell if it needs a `cdef`? I checked the generated C and it looks like Cython added a declaration for `alt`. Is that what you meant?

```
static CYTHON_INLINE int __pyx_f_5pysam_7libcbcf_has_symbolic_allele(struct __pyx_obj_5pysam_7libcbcf_VariantRecord *__pyx_v_record) {
  long __pyx_v_i;
  char *__pyx_v_alt;
  int __pyx_r;
  __Pyx_TraceDeclarations
  __Pyx_RefNannyDeclarations
  uint32_t __pyx_t_1;
  long __pyx_t_2;
  int __pyx_t_3;
  int __pyx_t_4;
  size_t __pyx_t_5;
```

bioinformed:
No worries.  I'll take a look at it tonight.

msto:
Hi Kevin, 

Were you able to take a look at this?

Thanks!

Matt

msto:
Thanks for the merge!

Issue #399, opened 2017-02-02
Allow use of '*' in region string


kghosesbg:
A very useful feature of `samtools view` is the `*` region tag that allows us to fetch all the unmapped reads in the file stuck at the end because they have no reference_id. It would be nice to have this in `psyam.fetch`. Currently the code complains `ValueError: invalid reference '*'`. Thanks so much!

AndreasHeger:
Thanks, I did not even know that was possible. Will add this for the next release.

byb121:
Hi,
I encountered the same issue, so just wondering what's the progress on this. 

Thanks,

AndreasHeger:
Thanks for reminding me, this has now been added.

dpryan79:
Great minds think alike? I assume that that'll work. Since it's in master now I'll just test that on a CRAM file to double check. I'll open a PR based on that should it be needed.

PR #573, opened 2017-11-17
Support the equivalent of HTS_IDX_NOCOOR when using fetch() or count()

dpryan79:
N.B., this isn't in a finished state yet, this is mostly meant as a starting point for discussion.

It's sometimes needed to be able to either see how many unmapped alignments with no assigned template are in a file and/or to iterate over them. In BAM files, these can at least be counted with idxstats, but since that doesn't work with CRAM files one need to `fetch(until_eof=True)` or `count(until_eof=True)`. If one is already using multiple threads to process regions (like we do in deepTools) then forcing this step to be single threaded adds a LOT of additional overhead. In HTSLIB, on can use the `HTS_IDX_NOCOOR` (-2) constant as a tid to jump straight to these entries. That would make things much more convenient for us in deepTools and perhaps for others that want to iterate over these reads in BAM or CRAM files.

The current implementation uses `tid = -2`, which is obviously a terrible implementation and would need to be changed before merging. However I'm hopeful that at least the concept of supporting HTS_IDX_NOCOOR is acceptable.

### DIFF
--- a/pysam/libcalignmentfile.pyx
+++ b/pysam/libcalignmentfile.pyx
@@ -737,7 +737,7 @@ cdef class AlignmentFile(HTSFile):
         backward compatiblity as synonyms for :term:`contig` and `stop`,
         respectively.
 
-        Without a `contig` or `region` all mapped reads in the file
+        Without a `contig`, `region`, or `tid` all mapped reads in the file
         will be fetched. The reads will be returned ordered by reference
         sequence, which will not necessarily be the order within the
         file. This mode of iteration still requires an index. If there is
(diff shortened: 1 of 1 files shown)


AndreasHeger:
Hi @dpryan79 , I think we have just worked at cross purposes, I worked on issue #399 and implemented something similar:
https://github.com/pysam-developers/pysam/pull/572

dpryan79:
Great minds think alike? I assume that that'll work. I'll install the master branch and test a CRAM file with it and open a new PR if needed. Thanks!

AndreasHeger:
Thanks!

PR #586, opened 2017-12-06
Allow loading index without tbi extension

mvdbeek:
This seems to be necessary to create TabixFile instances with indexes that don't end in `.tbi`.


bgruening:
:+1: looks good to me!

AndreasHeger:
Great, thanks!

PR #589, opened 2017-12-13
Return None from build_alignment_sequence if no MD tag is set

### LINKED ISSUE #588: get_reference_sequence should raise ValueError when no MD tag is present
Documentation for the `AlignedSegment` class indicates that the `get_reference_sequence` method requires an MD tag to be set, otherwise `ValueError` should be raised. The actual behaviour is that `get_reference_sequence` returns the query sequence (with gaps) when there is no MD tag.

The root cause is that `libcalignedsegment.build_alignment_sequence` is expected to return `None` when the MD tag is missing, but instead it returns the query sequence.

See [here](https://github.com/pysam-developers/pysam/blob/master/pysam/libcalignedsegment.pyx#L749) and [here](https://github.com/pysam-developers/pysam/blob/master/pysam/libcalignedsegment.pyx#L828).

j3harvey:
Closes #588.

Ensures that `AlignedSegment.get_reference_sequence` raises an exception when no MD tag is set.

### DIFF
--- a/pysam/libcalignedsegment.pyx
+++ b/pysam/libcalignedsegment.pyx
@@ -689,6 +689,10 @@ cdef inline bytes build_alignment_sequence(bam1_t * src):
     if src == NULL:
         return None
 
+    cdef uint8_t * md_tag_ptr = bam_aux_get(src, "MD")
+    if md_tag_ptr == NULL:
+        return None
+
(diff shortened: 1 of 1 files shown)


AndreasHeger:
Thanks!

PR #594, opened 2017-12-22
{AH} permit Number=G in INFO records

AndreasHeger:
Removing the assertion seems to enable the functionality without any side effects as far as I can see.

Would appreciate some comments though as I don't fully understand that part of the htslib API.

### DIFF
--- a/pysam/libcbcf.pyx
+++ b/pysam/libcbcf.pyx
@@ -222,8 +222,8 @@ cdef inline int is_gt_fmt(bcf_hdr_t *hdr, int fmt_id):
 
 
 cdef inline int bcf_genotype_count(bcf_hdr_t *hdr, bcf1_t *rec, int sample) except -1:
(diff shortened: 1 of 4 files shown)


bioinformed:
I'm pretty sure this issue is about Number=G in *INFO* fields, not FORMAT.  This was made illegal because FORMAT=G requires knowledge of ploidy.  I've filed an issue in hts specs requesting clarification.  See https://github.com/samtools/hts-specs/issues/272

AndreasHeger:
Sorry, yes, I meant INFO. 

Thanks for raising the issue, looks it is being clarified.

AndreasHeger:
I have reverted the change ensuring that the Gnomad file in its current form with `Number=G` will fail. Added a test case to ensure that the workaround replacing it with `Number=.` works. 
The PR now only adds the tests but leaves pysam behaviour unchanged.

Issue #579, opened 2017-11-21
Add ability to specify multiple compression threads when writing bam/cram


keiranmraine:
I'm aware that HTSlib supports a thread pool for use when compressing the output stream during generation of BAM/CRAM files.  Could this be exposed?

bioinformed:
Absolutely-- this has been on my TODO list for some time.

AndreasHeger:
Looks like this is not too difficult, shall I have a go (as an addition to HtsFile._open_htsfile)?

The question might be if we want to expose the threadpool interface of htslib so that the pool could be shared between multiple files. I think though it might be good to keep it simple and just provide the user with a "threads" (int) argument.

mvdbeek:
Writing can benefit quite strongly from using `hts_set_threads`:
```
In [49]: def benchmark_write(set_threads=False):
    ...:     source = pysam.AlignmentFile('/Users/mvandenb/src/readtagger/tests/h4.bam')
    ...:     out = pysam.AlignmentFile('out.bam', mode='wb', template=source)
    ...:     if set_threads:
    ...:         source.set_threads()
    ...:         out.set_threads()
    ...:     for _ in range(1000000):
    ...:         out.write(next(source))
    ...:
    ...:

In [50]: %timeit benchmark_write(set_threads=True)
1 loop, best of 3: 2.67 s per loop

In [51]: %timeit benchmark_write(set_threads=False)
1 loop, best of 3: 11.4 s per loop
```
This is when using 4 threads, that's pretty much the optimum we could expect
I'll clean this up and PR it.

bioinformed:
Thanks!  I'm happy to review such a PR.  I've been meaning to implement this for some time.

PR #599, opened 2018-01-04
Add a format_options keyword to AlignmentFile

dpryan79:
This implements #579 and additionally allows decoding specific fields in CRAM files (in the same was as `--input-fmt-option` and `--output-fmt-option` in `samtools view`).

Another possible implementation of this would be to use a `dict()` (I'm not sure what people will find more convenient).

N.B., it's inadvisable to use `required_fields=XXX` and then to attempt to use one of the unrequired fields (try it, you'll get some "interesting" results that I presume are an artefact from htslib).

### DIFF
--- a/pysam/libcalignmentfile.pyx
+++ b/pysam/libcalignmentfile.pyx
@@ -717,6 +717,10 @@ cdef class AlignmentFile(HTSFile):
         Issue a warning, instead of raising an error if the current file
         appears to be truncated due to a missing EOF marker.  Only applies
         to bgzipped formats. (Default=False)
+
+    format_options: list
+        A list of key=value strings, as accepted by --input-fmt-option and
+        --output-fmt-option in samtools.
     """
 
     def __cinit__(self, *args, **kwargs):
@@ -787,7 +791,8 @@ cdef class AlignmentFile(HTSFile):
               referencenames=None,
               referencelengths=None,
               duplicate_filehandle=True,
-              ignore_truncation=False):
+              ignore_truncation=False,
+              format_options=None):
         '''open a sam, bam or cram formatted file.
 
         If _open is called on an existing file, the current file
@@ -892,6 +897,8 @@ cdef class AlignmentFile(HTSFile):
                                   force_str(strerror(errno))))
                 else:
                     raise ValueError("could not open alignment file `{}`".format(force_str(filename)))
+            if len(format_options):
+                self.add_hts_options(format_options)
             # set filename with reference sequences. If no filename
             # is given, the CRAM reference arrays will be built from
             # the @SQ header in the header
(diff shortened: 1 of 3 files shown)


pysam/libchtslib.pyx
```diff
@@ -562,6 +562,22 @@ cdef class HTSFile(object):
...
+    def add_hts_options(self, format_options=[]):
+        """Given a list of key=value format option strings, add them to an open htsFile
+        """
+        cdef int rval
+        cdef hts_opt *opts = NULL
+
+        for format_option in format_options:
+            rval = hts_opt_add(&opts, format_option)
+            if rval != 0:
+                raise AttributeError('Invalid format option ({}) specified'.format(format_option))
```
bioinformed: free `opts`?

pysam/libchtslib.pyx
```diff
@@ -562,6 +562,22 @@ cdef class HTSFile(object):
...
+        if opts != NULL:
+            rval = hts_opt_apply(self.htsfile, opts)
+            if rval != 0:
+                raise AttributeError('An error occured while applying the requested format options')
```
bioinformed: free `opts`?

pysam/libchtslib.pyx
dpryan79: Yup, good catch

pysam/libchtslib.pyx
dpryan79: Yup, good catch

dpryan79:
I think I need to change the default to `[]` from `None`, so in the off-chance that this actually passes testing please don't merge yet!

dpryan79:
OK, those are now fixed and tested locally.

pysam/libcalignmentfile.pyx
```diff
@@ -787,7 +791,8 @@ cdef class AlignmentFile(HTSFile):
               referencenames=None,
               referencelengths=None,
               duplicate_filehandle=True,
-              ignore_truncation=False):
+              ignore_truncation=False,
+              format_options=[]):
```
bioinformed: Why is it better to use `[]` than `None`?  It is generally bad practice to use mutable values as default values and I didn't see any problems with using `None` as before.

dpryan79:
It’s purely convenience, since then one can just iterate over the list. I can change this to handle None easily enough though. I’ll change that later today.

dpryan79:
This PR should be good to go now.

AndreasHeger:
Thanks, brilliant

dpryan79:
Cool, thanks!

Relatedly, I just discovered that not requiring the sequence in a CRAM file affects the alignments returned by `fetch()`. This happens in `samtools view` too, so it's an htslib issue that I'll report there.

Issue #603, opened 2018-01-16
TypeError: "NoneType"  when running count_coverage


nickschurch:
When I load a bam file and call `count_coverage` on a region I'm getting an none error:

```
> thisbam = pysam.AlignmentFile(bam1,"rb")
> thisbam.text
'@HD\tVN:1.3\tSO:coordinate\n@SQ\tSN:1\tLN:30427671\n@SQ\tSN:2\tLN:19698289\n@SQ\tSN:3\tLN:23459830\n@SQ\tSN:4\tLN:18585056\n@SQ\tSN:5\tLN:26975502\n@SQ\tSN:Mt\tLN:366924\n@SQ\tSN:Pt\tLN:154478\n'
> thisbam.count("1",10,100000)
1724
> thisbam.count_coverage("1",10,100000)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-44-5c08c453e230> in <module>()
----> 1 thisbam.count_coverage("1",10,100000)

pysam/libcalignmentfile.pyx in pysam.libcalignmentfile.AlignmentFile.count_coverage()

TypeError: 'NoneType' object has no attribute '__getitem__'
```
Python 2.7.13 running pysam 0.13 on centos 6.3

AndreasHeger:
Thanks. Is it possible to share some sample data where this error occurs - I have trouble reproducing it.

mv911119:
[af.zip](https://github.com/pysam-developers/pysam/files/1699163/af.zip)
Here I have the same problem and this is my 16S data.
the chromosomes' names are 0,1,2,3,4.....
And Nick used Python 2 which could be the problem.
But I am using Python 3.6.2 and Centos Linux 7 And having the same in chromosome region reading

import pysam
sbm1 = pysam.AlignmentFile("/afsecondfiltered.bam","rb")
sbm1.count_coverage(contig="1", start= 1, stop=12)
Traceback (most recent call last):

  File "<ipython-input-68-2e971b60de51>", line 3, in <module>
    sbm1.count_coverage(contig="1", start= 1, stop=12)

  File "pysam/libcalignmentfile.pyx", line 1235, in pysam.libcalignmentfile.AlignmentFile.count_coverage

TypeError: 'NoneType' object is not subscriptable

zhoudreames:
I also get this problem ,do you know how to solve it ? thanks
Traceback (most recent call last):
[...]

PR #621, opened 2018-02-12
Adding a warning to count_coverage when an alignment has an empty QUAL field

wckdouglas:
Trying to fix issue [#603](https://github.com/pysam-developers/pysam/issues/603). In the ```AlignmentFile.count_coverage()``` method, if an alignment has an empty QUAL field, it raised a ```TypeError```.

I am fixing this by ignoring the quality threshold and counting every bases when an alignment has empty QUAL field. I am adding a warning for these kind of ```AlignedSegment```, so this issue is not silently passed. However, I am not sure if a warning is the best solution or if it should just raise a more clear error?

### DIFF
--- a/pysam/libcalignmentfile.pyx
+++ b/pysam/libcalignmentfile.pyx
@@ -1572,7 +1572,7 @@ cdef class AlignmentFile(HTSFile):
                    _start <= refpos < _stop:
 
                     # only check base quality if _threshold > 0
-                    if (_threshold and quality and quality[qpos] > _threshold) or not _threshold:
+                    if (_threshold and quality and quality[qpos] >= _threshold) or not _threshold:
                         if seq[qpos] == 'A':
                             count_a.data.as_ulongs[refpos - _start] += 1
                         if seq[qpos] == 'C':
--- a/pysam/libcalignmentfile.pyx
+++ b/pysam/libcalignmentfile.pyx
@@ -1572,7 +1572,7 @@ cdef class AlignmentFile(HTSFile):
                    _start <= refpos < _stop:
 
                     # only check base quality if _threshold > 0
-                    if (_threshold and quality and quality[qpos] < _threshold) or not _threshold:
+                    if (_threshold and quality and quality[qpos] > _threshold) or not _threshold:
                         if seq[qpos] == 'A':
                             count_a.data.as_ulongs[refpos - _start] += 1
                         if seq[qpos] == 'C':
--- a/pysam/libcalignmentfile.pyx
+++ b/pysam/libcalignmentfile.pyx
@@ -1537,6 +1537,9 @@ cdef class AlignmentFile(HTSFile):
         cdef int refpos
         cdef int c = 0
         cdef int filter_method = 0
+
+        cdef int not_check_qual = 0
+
         if read_callback == "all":
             filter_method = 1
         elif read_callback == "nofilter":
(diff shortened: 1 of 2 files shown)


bioinformed:
I'd prefer that QUAL=None only pass if the mapq thresholding is disabled (i.e. set to None or 0).  This seems less magical and doesn't require a warning.

pysam/libcalignmentfile.pyx
```diff
@@ -1568,15 +1569,15 @@ cdef class AlignmentFile(HTSFile):
             seq = read.seq
             quality = read.query_qualities
             
-            not_check_qual = 0
-            if not quality:
-                warnings.warn('%s contains QUAL field, all bases were counted' %read.query_name)
-                not_check_qual = 1
+            if not quality and check_qual == 1:
+                raise ValueError('%s contains no QUAL field, use quality_threshold = None to ignore base quality' %read.query_name)
```
bioinformed: I do not recommend raising an error here.  If quality is not set, then it should fail the quality threshold check.

pysam/libcalignmentfile.pyx
```diff
@@ -1538,14 +1538,15 @@ cdef class AlignmentFile(HTSFile):
...
+        # check base quality if quality == None or 0
+        cdef int check_qual = 1 if quality_threshold else 0 
 
         if read_callback == "all":
             filter_method = 1
         elif read_callback == "nofilter":
             filter_method = 2
 
-        cdef int _threshold = quality_threshold
+        cdef int _threshold = quality_threshold if quality_threshold else 0
```
bioinformed: should work with `quality_threshold or 0`

pysam/libcalignmentfile.pyx
```diff
@@ -1538,14 +1538,15 @@ cdef class AlignmentFile(HTSFile):
         cdef int c = 0
         cdef int filter_method = 0
 
-        cdef int not_check_qual = 0
+        # check base quality if quality == None or 0
+        cdef int check_qual = 1 if quality_threshold else 0
```
bioinformed: `cdef bint check_qual = bool(quality_threshold)`

pysam/libcalignmentfile.pyx
wckdouglas: thanks! does that mean setting ```quality_threshold = None``` should not be accepted and should raise an error?

pysam/libcalignmentfile.pyx
wckdouglas: The reason I added this is that the error raised from failing the quality threshold check isn't so intuitive, since it is just an ```TypeError: 'NoneType' object has no attribute '__getitem__'```  from failing to excute ```quality[qpos]```. Is there a good way to fix this?

pysam/libcalignmentfile.pyx
bioinformed: No.  The intent of `None` is to have no quality threshold (basically the same as zero).
```python
>>> None or 0
0
```

wckdouglas: so is the current line acceptable?  I think the if-clause is needed here to manually set ```None``` to 0, otherwise, ```cdef int _threshold = None``` will raise a TypeError.  And I also have a question on this line, I have checked ```_threshold``` is not used in any other place in the function, so I am guessing this line is just for making sure ```quality_threshold``` is set to an integer, right?

bioinformed: Does this not work?  It is more succinct and should handle `None`s properly. 
```python
    cdef int _threshold = quality_threshold or 0
```

pysam/libcalignmentfile.pyx
bioinformed: I'd write the check to be:
```python
    if _threshold and (not quality or quality[qpos] < _threshold):
```

pysam/libcalignmentfile.pyx
wckdouglas: it works! no idea there was this trick. thanks!

AndreasHeger:
Thanks!

PR #635, opened 2018-03-02
Speed up find intron

wckdouglas:
The current implementation of ```AlignmentFile.find_intons``` turns out to be very inefficient. The current function relies on ```AlignedSegment.get_aligned_pairs```, which can be extremely slow when it encounters long range splice junctionas, since it [loops over the every base on the read alignments and every position along the genome](https://github.com/pysam-developers/pysam/blob/master/pysam/libcalignmentfile.pyx#L1582). 

As ```AlignedSegment.get_aligned_pairs``` relies on cigar string,```AlignmentFile.find_intons``` should not need to parse every position again to recalculate the splice junctions. In this pull request, I am fixing this inefficient loop by just parsing the cigar string. A benchmark test is as below:
  

```
Python 3.6.4 | packaged by conda-forge | (default, Dec 23 2017, 16:31:06) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.2.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import pysam
   ...: import timeit
   ...: bam = pysam.AlignmentFile('pysam/tests/pysam_data/ex_spliced.bam')
   ...: 
   ...: %timeit bam.find_introns(bam.fetch())
   ...: %timeit bam.find_introns_slow(bam.fetch())
   ...: 
   ...: new = bam.find_introns(bam.fetch())
   ...: old = bam.find_introns_slow(bam.fetch())
   ...: 
   ...: new == old
   ...: 
2.44 ms ± 1.08 ms per loop (mean ± std. dev. of 7 runs, 1000 loops each)
6.26 ms ± 1.82 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
Out[1]: True
```


AndreasHeger:
Great, thanks!

PR #638, opened 2018-03-06
Allow setting number of threads for htslib

mvdbeek:
Internally this uses hts_set_threads. There is only a minor
speedup for reading files, but writing files can benefit well
from setting n_threads to more than 1:

```
In [1]: def benchmark_write(threads=1):
   ...:     source = pysam.AlignmentFile('/Users/mvandenb/src/readtagger/tests/h4.bam', threads=threads)
   ...:     out = pysam.AlignmentFile('out.bam', mode='wb', template=source, threads=threads)
   ...:     for _ in range(1000000):
   ...:         out.write(next(source))
   ...:

In [2]: %timeit benchmark_write(threads=1)
1 loop, best of 3: 9.21 s per loop

In [3]: %timeit benchmark_write(threads=2)
1 loop, best of 3: 4.68 s per loop

In [4]: %timeit benchmark_write(threads=3)
1 loop, best of 3: 3.17 s per loop

In [5]: %timeit benchmark_write(threads=4)
1 loop, best of 3: 2.66 s per loop
```
This had been discussed in https://github.com/pysam-developers/pysam/issues/579#issuecomment-370766946.

### DIFF
--- a/pysam/libcalignmentfile.pyx
+++ b/pysam/libcalignmentfile.pyx
@@ -567,7 +567,7 @@ cdef class AlignmentFile(HTSFile):
     header=None, add_sq_text=False, check_header=True, check_sq=True,
     reference_filename=None, filename=None, index_filename=None,
     filepath_index=None, require_index=False, duplicate_filehandle=True,
-    ignore_truncation=False)
+    ignore_truncation=False, n_threads=1)
 
     A :term:`SAM`/:term:`BAM`/:term:`CRAM` formatted file.
 
@@ -711,6 +711,8 @@ cdef class AlignmentFile(HTSFile):
     format_options: list
         A list of key=value strings, as accepted by --input-fmt-option and
         --output-fmt-option in samtools.
+    n_threads: integer
+        Number of threads to use for compressing/decompressing BAM/CRAM files. (Default=1)
     """
 
     def __cinit__(self, *args, **kwargs):
@@ -782,7 +784,8 @@ cdef class AlignmentFile(HTSFile):
               referencelengths=None,
               duplicate_filehandle=True,
               ignore_truncation=False,
-              format_options=None):
+              format_options=None,
+              n_threads=1):
         '''open a sam, bam or cram formatted file.
 
         If _open is called on an existing file, the current file
@@ -886,6 +889,7 @@ cdef class AlignmentFile(HTSFile):
                                  "header, text or reference_names/reference_lengths")
             
             self.htsfile = self._open_htsfile()
+            hts_set_threads(self.htsfile, n_threads)
 
             if self.htsfile == NULL:
                 if errno:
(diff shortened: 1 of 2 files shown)


pysam/libcalignmentfile.pyx
```diff
@@ -992,6 +997,16 @@ cdef class AlignmentFile(HTSFile):
...
 
+        # Set any eventual extra threads
+        if n_threads > 1 and ignore_truncation:
+            # This won't raise errors if reaching a truncated alignment,
+            # because bgzf_mt_reader in htslib does not deal with
+            # bgzf_mt_read_block returning non-zero values, contrary
+            # to bgzf_read (https://github.com/samtools/htslib/blob/1.7/bgzf.c#L888)
+            # Better to avoid this (for now) than to produce seemingly correct results.
+            raise ValueError('Cannot add extra threads when "ignore_truncation" is True')
+        hts_set_threads(self.htsfile, n_threads - 1)
```
bioinformed: Ideally, this call would be part of the `HTSFile` base object, so that `VariantFile` and other objects also benefit.  I very much like the conservative approach to interactions with truncation detection.  Also, please rename `n_threads` to `threads`, since we don't use the `n_` prefix elsewhere in the code.

mvdbeek:
Just pushing out the new changes to see if things pass on Travis, I'll also add `threads` to VariantFile and TabixFile

mvdbeek:
Alright, I've added hts_set_threads to `HTSFile` and exposed a `threads` parameter in `VariantFile` and `TabixFile` as well.

AndreasHeger:
Thanks, is it ready to be merged?

bioinformed:
I've run some light tests and all appears to work.  I've been working on some CRAM threading issues, so I didn't get to do as much testing as I'd hoped.

mvdbeek:
If there's something I can look into or there are scenarios that need more test coverage let me know, I can work on that.

AndreasHeger:
Thanks both. I want to make a bugfix release, so might keep this PR out but move it right afterwards.

bioinformed:
p.s. my reason for delay is less about this PR, but about upstream CRAM threading fixes.  This feature will be much safer once those are present in pysam.

jkbonfield:
Probably wise!

For what it's worth, I'm fairly confident we nailed them.  I've been through some randomly generated CRAM files with differing numbers of seqs per slice, slices per containers, and allocation of block encoding methods using gcc thread sanitiser, valgrind drd and valgrind helgrind.  It looks OK, but it's still hard to cover every possible scenario.

bioinformed:
@jkbonfield: Thanks for the update.  Do you have any idea how imminent the htslib release will be?

bioinformed:
@jkbonfield
p.s. the CRAM issues I alluded to above involved testing the recent htslib changes in a large code base utilizing threading and CRAM.  The latest `develop` branch is passing all tests, though not all steps have threading enabled.  We'd see random intermittent crashes with htslib 1.7, so the current state is looking good!

mvdbeek:
Given that 0.14.1 is out for a while now is there any chance we can merge this ? I guess there'd be a higher chance to catch bugs if this is in some time before the next release is coming up ?

mvdbeek:
Thanks for maintaining pysam, much appreciated!

PR #647, opened 2018-03-27
Fix operator precedence error by removing premature optimistion

mp15:
!= has higher precedence than & and will be evaluated first.
Replaced with clearer expression.

### DIFF
--- a/import.py
+++ b/import.py
@@ -103,7 +103,7 @@ def _update_pysam_files(cf, destdir):
                 lines = re.sub("stdout", "{}_stdout".format(basename), lines)
                 lines = re.sub(" printf\(", " fprintf({}_stdout, ".format(basename), lines)
                 lines = re.sub("([^kf])puts\(([^)]+)\)",
-                               r"\1fputs(\2, {}_stdout) & fputc('\\n', {}_stdout)".format(basename, basename),
+                               r"\1fputs(\2, {}_stdout) != EOF && fputc('\\n', {}_stdout)".format(basename, basename),
                                lines)
                 lines = re.sub("putchar\(([^)]+)\)",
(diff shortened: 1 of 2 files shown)


kyleabeauchamp:
Isn't this file automatically generated from samtools bam.c?

See, e.g., https://github.com/pysam-developers/pysam/blob/master/import.py#L89

Thus, we might not want to manually edit this?

jmarshall:
Indeed. This bug is being copied from [import.py:_update_pysam_files()](https://github.com/pysam-developers/pysam/blob/09fb70ed9f6281db45a27bde2c631a34920dd862/import.py#L105-L107), so you Pythonites will probably want to fix it at source. (That doesn't look fragile at all!)

kyleabeauchamp:
If you have easy suggestions on how to streamline the HTSLib <-> pysam integration, I'm all ears ;)

mp15:
Try this updated version then, != has higher precedence than && so it should be safe and get rid of the error.

AndreasHeger:
Many thanks for bringing this up, @mp15 
@kyleabeauchamp is right that this is an upstream issue. 
The aim of import.py is to only make changes that are absolutely necessary
to be able to link in samtools/bcftools.

AndreasHeger:
Closing PR

jmarshall:
You realise that “upstream” is pysam's own _devtools/import.py_ and that it's still broken, right?

kyleabeauchamp:
Sadly I think you'll have to re-run the HTSLib upgrade after you merge this :(

AndreasHeger:
thanks, @kyleabeauchamp

kyleabeauchamp:
Thanks @mp15, we've fixed this using the approach of @jmarshall (#706, #707).  Thanks for the bug report and your patch!

PR #662, opened 2018-04-07
Move lots of stuff from top-level dir to /devtools

kyleabeauchamp:
1.  Renamed /ci as /devtools
2.  Moved lots of "occasional" scripts from the top-level dir to /devtools

I *think* this makes it easier for folks to navigate the source tree and find things, but we can debate whether agree or not.


AndreasHeger:
Good idea, I am in favour of this.

Issue #669, opened 2018-04-20
Expose bcf_get_value_count on the record object


TMiguelT:
`bcf_get_value_count` seems to calculate how many values a given INFO/FMT field should have, on a given record. This is a useful feature, if you want to edit a VCF record but don't know how many values each field should have. Could this be added to the `VariantRecord` object, e.g. as `VariantRecord#get_value_count(type: str, id: str)`?

I'll probably try writing this myself at some point, but I wanted to keep track of this as an issue

bioinformed:
Good idea!  It will be a few weeks before I have time to work on this, but it will be extremely useful.

TMiguelT:
Any chance of merging this?

TMiguelT:
Totally agree with the interface of putting `value_count` inside the sample/info object. I forgot that they were actually C objects and not just dictionaries, and so can have methods.

I'm not sure I follow in terms of `GT` though. It's a field that always has a Number of R, effectively (although `|` or `/` separated, rather than comma separated). You can calculate the length of the field with `len(gt)`, but I see no reason not to allow this function to cover all fields in a general way.

PR #672, opened 2018-04-22
Implement VariantRecord#calc_number

multimeric:
See #669

- Expose bcf_get_value_count as VariantRecord#calc_number
- Add tests for this
- Some extra docstrings

### DIFF
--- a/pysam/libcbcf.pyx
+++ b/pysam/libcbcf.pyx
@@ -100,7 +100,6 @@ from cpython.version cimport PY_MAJOR_VERSION
 
 from pysam.libchtslib cimport HTSFile, hisremote
 
-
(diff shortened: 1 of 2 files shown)


TMiguelT:
Some questions about my implementation:
* My editor auto-formatted the code, should I undo that?
* I make users specify `INFO` or `FMT` as a string to determine the type of field. Should I expose an enum/constant for these values instead? If strings, should I use `FORMAT` instead of `FMT`? Do we have a precedent?

bioinformed:
Re editor reformatting: Yes-- you'll need to undo all of that for this MR to be considered.
Re API for FMT vs. INFO: I left some suggestions in #669.

PR #675, opened 2018-05-09
Documentation typo

ajuric:
Matching the documentation from samtools mpileup and get_query_sequences() method.


AndreasHeger:
Thanks!

PR #694, opened 2018-07-02
Ensure that = and X CIGAR ops are treated the same as M

dpryan79:
It turns out that if you try to use the `get_blocks()` function on a read with CIGAR string `150=` you'll get `[]` returned. This is due to a number of functions not handling the `=` and `X` operations. I've fixed that in the current PR.


dpryan79:
If you want a test dataset you can see item 5 in [this Galaxy history](https://usegalaxy.eu/u/helena-rasche/h/unnamed-history-1).

AndreasHeger:
Many thanks!

PR #705, opened 2018-07-24
[WIP] Re-run import.py with changes from #647

kyleabeauchamp:
To get the ball rolling, I've made an updated version of #647 that applies the `import.py` transformation to all of the vendored files in pysam.  

I ran the following commands:

```
tar -jxvf htslib-1.9.tar.bz2

rm -rf htslib
mv htslib-1.9 htslib
echo '#define HTS_VERSION "1.9"' > htslib/version.h

tar -jxvf samtools-1.9.tar.bz2


rm -rf samtools
python devtools/import.py samtools samtools-1.9/
echo '#define SAMTOOLS_VERSION "1.9"' > samtools/version.h

tar -jxvf bcftools-1.9.tar.bz2

rm -rf bcftools
python devtools/import.py bcftools bcftools-1.9
echo '#define BCFTOOLS_VERSION "1.9"' > bcftools/version.h

# Manually edit pysam/version.py
# Revert a couple of changed files to master
git checkout htslib/htslib.pc.tmp
git checkout bcftools/config.h

# Manually comment out two problematic bam_tview_main lines in samtools/bamtk.c.pysam.c

python setup.py install
```

### DIFF
--- a/bcftools/tabix.c.pysam.c
+++ b/bcftools/tabix.c.pysam.c
@@ -78,7 +78,7 @@ int main_tabix(int argc, char *argv[])
         BGZF *fp;
         s.l = s.m = 0; s.s = 0;
         fp = bgzf_open(argv[optind], "r");
-        while (bgzf_getline(fp, '\n', &s) >= 0) fputs(s.s, bcftools_stdout) & fputc('\n', bcftools_stdout);
+        while (bgzf_getline(fp, '\n', &s) >= 0) fputs(s.s, bcftools_stdout) != EOF && fputc('\n', bcftools_stdout);
         bgzf_close(fp);
         free(s.s);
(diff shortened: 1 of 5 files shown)


kyleabeauchamp:
I haven't thought carefully about whether we need the extra EOF comparison or not.  Nor have I looked at whether this impacts the tests.

kyleabeauchamp:
see also #647

kyleabeauchamp:
OK, looks like Travis passes.  Still need a review.

jmarshall:
> I haven't thought carefully about whether we need the extra EOF comparison or not.

You do, but really this shows what a bad idea it is to try to rewrite this stuff to two function calls when you don't know what the surrounding code looks like.

I'd recommend the approach in #706 to fix this tiny problem instead.

PR #706, opened 2018-07-25
Really fix this import.py fputs != EOF && fputc != EOF madness

jmarshall:
Rather than trying to rewrite a function call as two function calls when we don't really know what the syntax around the call site looks like, rewrite it as a call to our own function.

I haven't actually run this, but assuming the regexp syntax is right you should get the idea. Needs a followup commit to actually run _import.py_ over the freshly imported source code.


jmarshall:
Supercedes #647 and the import.py changes in #705.

kyleabeauchamp:
Thanks @jmarshall , I applied your patch in #707 and regenerated the wrapper C files.

PR #715, opened 2018-07-29
pysam_data: cleaning of generated with_md.sam.gz

smoe:
It is just something one notices when packaging for Debian.


AndreasHeger:
Thanks

PR #737, opened 2018-10-26
Fix bug in get_aligned_pairs

bkohrn:
I  noticed that in pysam 0.15.1, there is a bug with the get_aligned_pairs function when matches_only=True and with_seq=True.  Basically, the read position gets advanced, but the reference position does not.  This fixes that bug, and I believe it won't involve changing any other portions of the code.

### DIFF
--- a/pysam/libcalignedsegment.pyx
+++ b/pysam/libcalignedsegment.pyx
@@ -1952,6 +1952,8 @@ cdef class AlignedSegment:
                     else:
                         for i from pos <= i < pos + l:
                             result.append((None, i))
+                else:
+                    r_idx += l
                 pos += l
 
             elif op == BAM_CHARD_CLIP:


AndreasHeger:
Thanks, apologies for getting back so late with this PR.

I think r_idx, the pointer to the current position in the reference sequence, is only used if with_seq = True, so does not need to be incremented if there is no reference sequence. Basically, r_idx should be incremented after every time it is referenced:

bkohrn:
I think you misunderstand the arguments...from the documentation (https://pysam.readthedocs.io/en/latest/api.html):  

get_aligned_pairs(self, matches_only=False, with_seq=False)
   a list of aligned read (query) and reference positions.  
   For inserts, deletions, skipping either query or reference position may be None.
   Padding is currently not supported and leads to an exception.
   Parameters:   
      matches_only (bool) - If True, only matched bases are returned - no None on either side.
      with_seq (bool) – If True, return a third element in the tuple containing the reference sequence. Substitutions are lower-case. This option requires an MD tag to be present.
   Returns: aligned_pairs
   Return type: list of tuples

I can show this using the following code on the attached file (contains a single read, which has a deletion):  
[Test.bam.zip](https://github.com/pysam-developers/pysam/files/2738648/Test.bam.zip)

```
import pysam
inBam = pysam.AlignmentFile("Test.bam" ,'rb')
for read in inBam:
    if read.has_tag("MD"):
        if "D" in read.cigarstring:
            readMD1 = read.get_aligned_pairs(matches_only=True, with_seq=True)
            readMD2 = read.get_aligned_pairs(matches_only=False, with_seq=True)
            print("read.get_aligned_pairs(matches_only=True, with_seq=True)")
            print(readMD1)
            print()
            print("read.get_aligned_pairs(matches_only=False, with_seq=True)")
            print(readMD2)
```

This excel file has the expanded form of the output, which should be easier to read.  As you can see, although the query positions are filtered correctly, the reference positions (and thus sequence) are not.  
[pysam_problem.xlsx](https://github.com/pysam-developers/pysam/files/2738654/pysam_problem.xlsx)

AndreasHeger:
Thanks for explaining, I see the issue now. I will build a test case and apply the fix.

PR #743, opened 2018-11-16
Update api.rst

liyao001:
When the window is a little bit larger, the original demo may raise error `exit code 139 (interrupted by signal 11: SIGSEGV)`. One assumed reason for this is the garbage collector in python deletes bamfile.pileup (...) while the loop is running and the code goes to empty memory. And to address this problem, the updated demo should be used.
Ref: https://sites.google.com/a/epigenetic.ru/lab-members/kb/it/pysam

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -57,7 +57,10 @@ reads are represented as :class:`~pysam.PileupRead` objects in the
 
     import pysam
     samfile = pysam.AlignmentFile("ex1.bam", "rb" )
-    for pileupcolumn in samfile.pileup("chr1", 100, 120):
+    
+    pileup = bamfile.pileup("chr1", 100, 120)
+
+    for pileupcolumn in pileup:
         print ("\ncoverage at base %s = %s" %
                (pileupcolumn.pos, pileupcolumn.n))
         for pileupread in pileupcolumn.pileups:


AndreasHeger:
Thanks @liyao001 - looks like an issue to me. Do have specific numbers (depth, size of window)? I am not sure the garbage collector will delete an object that is still referenced. Given this happens with large data, I am guessing memory is being allocated, but not checked if allocation was successful.

doc/api.rst
```diff
@@ -57,7 +57,10 @@ reads are represented as :class:`~pysam.PileupRead` objects in the
 
     import pysam
     samfile = pysam.AlignmentFile("ex1.bam", "rb" )
-    for pileupcolumn in samfile.pileup("chr1", 100, 120):
+    
+    pileup = bamfile.pileup("chr1", 100, 120)
```
ExplodingCabbage: This surely isn't what you meant. After the change, you're creating a variable called `samfile` above, but referencing `bamfile` here.

ExplodingCabbage:
After a quick glance, I can't make sense of why this change would alter garbage collector behaviour. It *is* possible for an iterable to be deleted before the end of a `for` loop that's iterating over it, but only if the iterator its `__iter__` returns doesn't hold a reference to the iterable. Example:

```python
class SomeIterable:
    def __del__(self):
        print('Iterable getting deleted')

    def __iter__(self):
        return iter(range(10))

print('Without storing in a variable...')
for el in SomeIterable():
    print(el)

print()

print('With storing in a variable...')
some_iterable = SomeIterable()
for el in some_iterable:
    print(el)
```

However, it looks to me like `pileup()` returns either an `IteratorColumn` or an `IteratorColumnRegion` (which inherits from `IteratorColumn`, thus calling `IteratorColumn.__cinit__` on construction because superclass `__cinit__` constructors are implicitly called in Cython). In `IteratorColumn.__cinit__` we have `self.samfile = samfile`. So there should be a reference to the samfile until the for loop ends. It seems impossible, therefore, for this change to affect garbage collection behaviour (or indeed fix anything in any way).

Am I missing something? @liyao001, are you able to provide a test case demonstrating the error?

PR #755, opened 2019-01-18
Allow passing flags and arguments to index methods

mvdbeek:


AndreasHeger:
Thanks, I once thought mapping all input arguments to python keyword arguments, but never got to it.

PR #756, opened 2019-01-26
Add AlignedSegment.compare_sort

mvdbeek:
This allows users to compare reads via coordinate, queryname or tag
using the samtools rules.
This should be significantly more performant (and easier) than
https://github.com/10XGenomics/supernova/blob/master/tenkit/lib/python/tenkit/bam.py#L975

### DIFF
--- a/devtools/import.py
+++ b/devtools/import.py
@@ -70,6 +70,38 @@ MAIN = {
 }
 
 
+# The follwing fragment adds an additional function that needs to go with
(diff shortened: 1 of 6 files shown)


mvdbeek:
I've implemented this now as https://github.com/mvdbeek/pysam-compare-reads, I guess there's no need to have this in pysam proper, but feel free to merge it anyway if you consider it useful.

PR #777, opened 2019-03-23
Add `with_ref` and `with_query` args for `get_aligned_pairs`.

yech1990:
- `with_seq` argument is misleading.
- add `with_query` will make pysam more powerful.


AndreasHeger:
Thanks, @yech1990 , I agree that with_seq is not a fortunate name and with_ref is better.

y9c:
More related issues 

https://github.com/pysam-developers/pysam/pull/1294

PR #780, opened 2019-03-27
_pysam_dispatch: close all files before opening new ones

### LINKED ISSUE #745: bam2fq crashes on third call
```
@HD	VN:1.4	SO:queryname
@RG	ID:A	SM:/Users/xxxx/working/test.fastq
M02781:22:000000000-A84MN:1:1101:10000:23637	4	*	0	0	*	*	0	0	GGTTTGACGCTCTTCCGATCTCACTTGAGCACTGCCATCAAGCTCAGGCTCTATTTTCTTAGCCAGCTGCTTTCCCAGCTCCACTCCCCACTGGTCAAAGCTGTTGATGTCCCAGATGATGCCCTGAACGAAGATCTTGTGCTCATACATG	BCCCCCFFFFCCGGGGGGGGGGHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHGHHHHHHHHHHHHHHHHHHHHHGHGHHHHHHHHHHHHHHHHHHHHHHHHHHHHHGHHHHHHHHHGGGGHHHHHHHHHHHHHHHHH	RG:Z:A
```

Convert this to bam:
`samtools view -b test.sam > test.bam`

Run pysam.bam2fq on this file a couple of times:
```
Python 2.7.5 (default, Jun 27 2014, 14:29:53)
[GCC 4.4.7 20120313 (Red Hat 4.4.7-4)] on linux2
>>> import pysam
>>> pysam.bam2fq('-n', '-0', 'test.fastq', 'test.bam')
''
>>> pysam.bam2fq('-n', '-0', 'test.fastq', 'test.bam')
''
>>> pysam.bam2fq('-n', '-0', 'test.fastq', 'test.bam')
Segmentation fault (core dumped)
```
This was identified as it seems to be always the third time that segfaults. Different files do not seem to make any difference to the crash.

This was seen on versions 15.0 and 15.1, corresponding to samtools/htslib/bcftools 1.7.

jmarshall:
Because `bcftools_`/`samtools_stderr` and `bcftools_`/`samtools_stdout` share the same file descriptors (they call `fdopen()` on the same fds) and because the underlying samtools/bcftools code will sometimes close its standard output itself (e.g. samtools bam2fq), after repeated invocations the existing `@pysam@_set`/`unset_stdout` functions eventually get out of sync and reuse their own file descriptors.

For example, issue #745 is due to `samtools_set_stdout()`'s previous _samtools_stdout_ referencing a stale fd number that is the same fd number as the newly-`mkstemp()`ed fd it is being updated to. The stale fd is then closed via `fclose()` so the `fdopen()` fails.

Instead of reopening the four FILE pointers to /dev/null **interleaved** with closing them, close all four in sequence and only once they are all closed (and thus do not reference stale fd numbers) reopen them — on the next `_pysam_dispatch()` invocation. Fixes #745.


jmarshall:
And, with that fourth commit, simplify matters by only opening and closing **two** FILE pointers.

PR #781, opened 2019-03-28
Default to multiple_iterators=False for pileups

### LINKED ISSUE #725: CRAM bug with .pileup
Calling `AlignmentFile.pileup('chr2')` on a CRAM does not properly fetch (it iterates over reads starting on chr1).  Converting the same CRAM to a BAM then calling `AlignmentFile.pileup('chr2')` works properly.

pysam 0.12.0.1 works fine, but >=0.14 is definitely not working

jmarshall:
For CRAM files, HTSlib iterators are in fact a stub struct — the real iterator limits are stored within the cram_fd itself, having been set via `cram_set_option(CRAM_OPT_RANGE)`. This means that reopening the CRAM file as a new cram_fd, which happens with `multiple_iterators=True`, is problematic as the iterator limits don't apply to the separate cram_fd.

Avoiding reopening the CRAM file fixes #725.

I was tempted to do something similar in `IteratorColumnAllRefs` but that one works as is (though I don't especially see why!). To be honest, it's not clear what `multiple_iterators=True` is for in general…

### DIFF
--- a/pysam/libcalignmentfile.pyx
+++ b/pysam/libcalignmentfile.pyx
@@ -2611,10 +2611,11 @@ cdef class IteratorColumnRegion(IteratorColumn):
                   int start = 0,
                   int stop = MAX_POS,
                   int truncate = False,
+                  int multiple_iterators = False,
                   **kwargs ):
 
         # initialize iterator
-        self._setup_iterator(tid, start, stop, 1)
+        self._setup_iterator(tid, start, stop, multiple_iterators)
         self.start = start
         self.stop = stop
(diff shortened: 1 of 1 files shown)


AndreasHeger:
Thanks, @jmarshall , the problem has been in the past that not all pysam users were fully familiar with what you can and can't do with files, especially that with a single file pointer you can't interleave multiple fetch() and pileup() calls on the same file. Pysam tried to hide this complexity, but this might not be possible any more or be too painful to initialize complex data structures.

PR #782, opened 2019-03-28
Fix travis builds

jmarshall:
Bump the samtools/htslib/bcftools versions used. Pinning 1.7 means Pysam testing is stuck forever at the broken-openssl/libcrypto condapocalypse.

At least, that's a theory on how to get around [these macOS failures](https://api.travis-ci.org/v3/job/512462692/log.txt):

```
dyld: Library not loaded: @rpath/libcrypto.1.0.0.dylib

  Referenced from: /Users/travis/miniconda3/envs/testenv/bin/samtools

  Reason: image not found

./devtools/run_tests_travis.sh: line 48:  2127 Abort trap: 6           samtools --version
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 os:
-  - linx
+  - linux
   - osx
 
 language: c
--- a/devtools/run_tests_travis.sh
+++ b/devtools/run_tests_travis.sh
@@ -38,15 +38,16 @@ conda config --add channels bioconda
 
 # pin versions, so that tests do not fail when pysam/htslib out of step
 # add htslib dependencies
-conda install -y "samtools=1.7" "bcftools=1.6" "htslib=1.7" xz curl bzip2
+conda install -y "samtools=1.9" "bcftools=1.9" "htslib=1.9" xz curl bzip2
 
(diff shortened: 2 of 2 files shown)


jmarshall:
That's differently broken, which is a form of progress…

jmarshall:
Uh.. I reckon that's worse than before! There's some more experimenting needed here.

kyleabeauchamp:
OK, do we need to revert this merge?  I think it's clear that the existing CI was broken in that it was combining unsupported versions.

kyleabeauchamp:
Reverted.  Please take another stab, hopefully with fixed tests :).

kyleabeauchamp:
And also sorry for the chaos, I mis-interpreted your above statement about progress and request for review as suggesting this was good to go as way towards fixing an existing broken CI.

jmarshall:
In retrospect I can see how the sarcasm dripping from my comment was not obvious 😢 — d'oh!

(The “request for review” was because it appears draft PRs don't get picked up by Travis…)

Actually please do merge your py3.7 PR first as the channel reordering may have an effect on what went wrong with this one.

PR #786, opened 2019-03-28
More travis fixes

kyleabeauchamp:
1.  Add py37 to Travis matrix
2.  Remove python 3.4 and 3.5 from Travis matrix, as they are no longer supported on Bioconda and conda-forge
3.  Clean up conda channel setup to match https://bioconda.github.io/
4.  Fix samtools version used by Travis
5.  Force ncurses to come from `conda-forge` channel to overcome various known issues.

This PR combines #782 and #784.  I wanted to kick off the CI to see if it uncovers anything.

PR #789, opened 2019-03-30
Fix Travis samtools etc versions and BAM comparisons

jmarshall:
1. Use `checkGZBinaryEqual` to compare BAM files, as the more recent samtools build likely uses libdeflate rather than zlib — which introduces spurious differences in the compressed streams.

2. Bump the samtools/htslib/bcftools versions used. Pinning 1.7 means Pysam testing is stuck forever at the broken-openssl/libcrypto condapocalypse.

3. Force ncurses to come from conda-forge channel to overcome various known issues.

4. Add py37 to Travis matrix

5. Remove python 3.4 and 3.5 from Travis matrix, as they are no longer supported on Bioconda and conda-forge

6. Clean up conda channel setup to match https://bioconda.github.io/

7. Use raw strings for regular expressions to avoid Python 3.7 "DeprecationWarning: invalid escape sequence".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 os:
-  - linx
+  - linux
   - osx
 
 language: c
@@ -7,9 +7,8 @@ language: c
 env:
   matrix:
     - CONDA_PY=2.7
-    - CONDA_PY=3.4
-    - CONDA_PY=3.5
     - CONDA_PY=3.6
+    - CONDA_PY=3.7
   global:
     - PYSAM_LINKING_TEST=1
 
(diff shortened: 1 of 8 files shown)


jmarshall:
Another attempt at #786 et al. Another commit to come (so don't merge this yet!) — I wanted to observe what travis does at this point as well.

jmarshall:
All tests passing; I think this is ready to be merged to sort out the Travis baseline.

kyleabeauchamp:
Thanks for brining this to the finish line.  I'm merging this now rather than waiting for the usual "release cycle mass merge" because I think it's important for the Travis tests to be behaving the way we want as people open PRs---and the existing CI script is clearly not doing the right thing.

jmarshall:
Yes, exactly. Now to rebase my other PRs…

PR #803, opened 2019-05-21
Fix docstring

### LINKED ISSUE #802: Docstring duplication error in get_forward_qualities
I believe the docstring of `AlignedSegment.get_forward_qualities` was copied from `get_forward_sequence` and is wrong:  https://github.com/pysam-developers/pysam/blob/master/pysam/libcalignedsegment.pyx#L1855

kyleabeauchamp:
fixes #802

### DIFF
--- a/pysam/libcalignedsegment.pyx
+++ b/pysam/libcalignedsegment.pyx
@@ -1853,7 +1853,7 @@ cdef class AlignedSegment:
         return s
 
     def get_forward_qualities(self):
-        """return the original read sequence.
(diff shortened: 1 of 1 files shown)


jmarshall:
Also the following sentence should talk about reversing qualities rather than reverse complementing reads.

kyleabeauchamp:
Thanks for the review, should be fixed now.

AndreasHeger:
Thanks @kyleabeauchamp and @jmarshall

PR #804, opened 2019-05-22
Remove string buffer size limit in get_query_sequences()

### LINKED ISSUE #717: How can I use get_query_sequences() when the average depth is more than 10000
when I used `pileupcolumn.get_query_sequences`  , I got a error : ` File "pysam/libcalignedsegment.pyx", line 3041, in pysam.libcalignedsegment.PileupColumn.get_query_sequences
AssertionError`

   I noticed `cdef uint32_t MAX_PILEUP_BUFFER_SIZE = 10000` ,  but at some positions in my bam, the depth may be more than it.  And I do not want to change the source code , because that will not work on my professor's cluster.
   What should I do ?

`pysam.__version__
'0.15.0' `

jmarshall:
Use HTSlib's `kstring_t`, which reallocates and enlarges its memory as needed, rather than a fixed-size char buffer. Fixes #717.

### DIFF
--- a/pysam/libcalignedsegment.pxd
+++ b/pysam/libcalignedsegment.pxd
@@ -70,7 +70,7 @@ cdef class PileupColumn:
     cdef int n_pu
     cdef AlignmentHeader header
     cdef uint32_t min_base_quality
-    cdef uint8_t * buf
+    cdef kstring_t buf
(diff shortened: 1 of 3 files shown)


AndreasHeger:
Thank you very much, apologies for the delay.

PR #805, opened 2019-05-23
Trivial .gitignore / repository cleanup

jmarshall:
Remove _htslib/htslib.pc.tmp_, which is a generated file and shouldn't be under source control (neither should _htslib/htslib-uninstalled.pc_ if it were to get built). Fix typo.


kyleabeauchamp:
LGTM, gunna merge this because it's trivial and let Andreas handle the PRs with actual substance.

PR #806, opened 2019-05-31
Fix extra quotes in docstrings

juliangehring:
Some docstrings start with 4 quote characters instead of 3 which then shows up as e.g.

> “bool indicating the current state of the file object.

in the [documentation](https://pysam.readthedocs.io/en/latest/api.html#pysam.FastxFile.closed). This PR remove the extra characters.


kyleabeauchamp:
LGTM, thanks!  Will merge after tests finish.

Issue #773, opened 2019-03-13
memory leak with  VariantRecord.id


cariaso:
```
import psutil
import sys
import os
from pysam import VariantFile

def main():

    process = psutil.Process(os.getpid())

    fn = '/usr/local/share/bcbio/genomes/Hsapiens/hg19/variation/dbsnp-151.vcf.gz'
    vcf_reader = VariantFile(fn)

    for rec in vcf_reader:
        # commenting out the next line to eliminate the memory leak
        rec.id
        mem = process.memory_info()[0]
        print(mem)

if __name__ == '__main__':
    main()
```

if rec.id is executed, memory bleeds away quickly.
If not, memory holds steady.
any VCF should be fine for your testing.

bioinformed:
Thanks for the report.  I'll look into it.

cariaso:
I see that the problem does not happen for all VCFs, but 
ftp://ftp.ncbi.nih.gov/snp/organisms/human_9606_b151_GRCh37p13/VCF/common_all_20180423.vcf.gz
size 1.5G is able to reproduce the memory leak.

cariaso:
fixed in https://github.com/cariaso/pysam/pull/1

PR #807, opened 2019-06-03
do not cache VariantRecord.id to avoid memory leak from pysam-developers#773

cariaso:
avoids the problem described at https://github.com/pysam-developers/pysam/issues/773


AndreasHeger:
Thanks, @cariaso , not caching seems to be a prudent choice, no matter how the leak arises. @bioinformed , would you agree?

bioinformed:
I agree — in retrospect it was a silly mistake to use a caching function in that property.

AndreasHeger:
Thanks

Issue #722, opened 2018-09-14
macOS and Linux wheels


pb-dseifert:
Hi @AndreasHeger 
would it be possible to get Linux and macOS tarballs for the 0.15.1 release of pysam? Thanks a lot!

AndreasHeger:
Certainly, will build these on the weekend.

AndreasHeger:
I have built the linux wheels, not sure at the moment why there is no python 3.7

wm75:
Hi @AndreasHeger and thanks for all the work!
Would love to see the 3.7 wheel. What is the issue with it?

AndreasHeger:
Thanks, the 3.7 wheel has been uploaded - turns out I had not updated the manywheels docker image for while...

wm75:
Well, that can happen :smile: Thanks for looking into this!

kyleabeauchamp:
I think this is resolved?

nsoranzo:
@kyleabeauchamp I can't find macOS wheels on https://pypi.org/project/pysam/ , so I'd say not completely.

kyleabeauchamp:
Oops, thanks!

nsoranzo:
@kyleabeauchamp @AndreasHeger  Would you be interested in automatic building of wheels and pushing to PyPI on release tagging?
I've recently worked on that for another Python library: https://github.com/bxlab/bx-python/blob/master/.travis.yml

AndreasHeger:
@nsoranzo , that would be very helpful. How do you want to proceed?

nsoranzo:
I can try to work on that in the next few days and open a PR. Someone having PyPI push access for pysam would then need to add their username and encrypted password to `.travis.yml` ( https://docs.travis-ci.com/user/environment-variables#defining-encrypted-variables-in-travisyml ). Sounds good?

AndreasHeger:
Excellent, thanks

nsoranzo:
This issue can be finally closed.

PR #814, opened 2019-06-22
Build wheels and upload them to PyPI

nsoranzo:
when tagging a release.

Build Linux and macOS binary wheels using cibuildwheel.
Build sdist with libcurl and s3 support.
Upload sdist and wheels to PyPI using twine.

xref. https://github.com/pysam-developers/pysam/issues/722

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,54 @@ env:
     - CONDA_PY=3.7
   global:
     - PYSAM_LINKING_TEST=1
+    - TWINE_USERNAME=
+    - secure: ''
+
+_deploy_common: &deploy_common
+  if: tag IS present
+  install:
+    - python3 -m pip install cibuildwheel twine
+
+matrix:
+  include:
+    - stage: deploy
+      os: linux
+      language: python
+      python: '3.5'
+      services:
+        - docker
+      env:
+        - CIBW_BEFORE_BUILD="yum install -y zlib-devel bzip2-devel xz-devel && pip install -r requirements.txt"
+        - CIBW_ENVIRONMENT='HTSLIB_CONFIGURE_OPTIONS="--disable-libcurl"'
+      addons:
+        apt:
+          packages:
+            - gcc
+            - g++
+            - libcurl4-openssl-dev  # for libcurl support in sdist
+            - libssl-dev  # for s3 support in sdist
+      <<: *deploy_common
+      script:
+        - set -e
+        - cibuildwheel --output-dir dist
+        - python3 -m pip install Cython
+        - python3 setup.py build_ext --inplace
+        - python3 setup.py sdist
+        - twine check dist/*
+        - twine upload --skip-existing dist/*
+    - stage: deploy
+      os: osx
+      language: generic
+      env:
+        - CIBW_BEFORE_BUILD="pip install -r requirements.txt"
+        - CIBW_ENVIRONMENT='HTSLIB_CONFIGURE_OPTIONS="--disable-libcurl"'
+      addons: {}
+      <<: *deploy_common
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -11,6 +11,54 @@ env:
     - CONDA_PY=3.7
   global:
     - PYSAM_LINKING_TEST=1
+    - TWINE_USERNAME=
```
nsoranzo: Here the PyPI username of one of the maintainers needs to be added.

.travis.yml
```diff
@@ -11,6 +11,54 @@ env:
...
+    - secure: ''
```
nsoranzo: Here the output of:
```
travis encrypt 'TWINE_PASSWORD=escaped_PyPI_password'
```
needs to be added. See https://docs.travis-ci.com/user/encryption-keys#note-on-escaping-certain-symbols for details.

AndreasHeger: Thanks, I will try to merge this this week.

AndreasHeger:
Thanks for the PR and also thanks for the instructions - I have just merged and added username and secret.

nsoranzo:
Cool, I look forward to the next (automated) release!

nsoranzo:
@AndreasHeger It looks like there's something wrong with the encryption of `TWINE_PASSWORD`, the deploy of v0.15.3 failed: https://travis-ci.org/pysam-developers/pysam

In the Travis logs I see a suspicious `export MY_SECRET_ENV=[secure]` .
Are you sure you ran

```console
travis encrypt 'TWINE_PASSWORD=escaped_PyPI_password'
```

and not `travis encrypt 'MY_SECRET_ENV=escaped_PyPI_password'` as in the example at https://docs.travis-ci.com/user/environment-variables/#encrypting-environment-variables ?

AndreasHeger:
That is quite possible..., let me check.

AndreasHeger:
Indeed, that is what I did - sorry! I have pushed a change. 
Is there a way to trigger the deployment for a particular build, or do we need to wait for the next release tag change?

.travis.yml
```diff
@@ -11,6 +11,54 @@ env:
...
+
+_deploy_common: &deploy_common
+  if: tag IS present
```
mvdbeek: I believe you should be able to deploy from master with
`if: branch = master AND type = push AND fork = false` based on https://github.com/travis-ci/travis-ci/issues/8767#issuecomment-344751779. You'd have to revert that after the successful deployment to limit to deployments on tag

nsoranzo:
Thanks for checking, `TWINE_PASSWORD` is exported now.  Unfortunately you need to tag a commit to enable the deploy, as I have currently implemented it.

.travis.yml
AndreasHeger: Thanks, @mvdbeek , I will give this a go.

nsoranzo: It worked, wheels have been uploaded to PyPI! You can now revert commit 0401ebed8cdce01b0f7da0b47fd41718b4a7e2ec .

AndreasHeger:
Hi @nsoranzo , there seems to be an issue with the wheels. I am trying to debug using the CIBW_TEST_COMMAND env var to see if we can catch issues before deploying. If you have any advice, that would be very much appreciated as I am not familiar with cibuildwheel.

nsoranzo:
@AndreasHeger Sorry for the trouble! I think commit abf896a52b53242c8649007c041d9393d5407360 is breaking the TravisCI configuration because tabs cannot be used for indentation.
`CIBW_TEST_COMMAND` seems the way to go, but I don't have experience on this.

PR #822, opened 2019-07-17
Include AlignmentHeader class in API documention

diekhans:
Add class AlignmentHeader to the API documentation.


AndreasHeger:
Thanks

PR #825, opened 2019-07-17
Ah fix cpy error

AndreasHeger:


kyleabeauchamp:
I'm not super familiar with that part of the code, but it seems reasonable to me.

kevinjacobs-progenity:
Testing now.  There may also be an ASCII assumption in the tabix code, if I remember correctly.

AndreasHeger:
Thanks both. Merging for now, seems to work.

PR #829, opened 2019-07-28
Run configure with the correct CC/CFLAGS/LDFLAGS env vars

### LINKED ISSUE #495: upgrade from pysam-0.9.1.2 to pysam-0.11.2.2 fails
Attached the full terminal log during upgrade  (triggered by installing CrossMap)

Thanks for your help, this error is now on two independent macs with different OS'es (Yosemite and ElCapitan). lzma.h seem to be missing!! but lzma is installed and the .h is present in my macport include folder

```
ll /opt/local/include/lzma.h 
-rw-r--r-- 1 root admin 9.6K Dec 31  2016 /opt/local/include/lzma.h
```

```
lzma --version
xz (XZ Utils) 5.2.3
liblzma 5.2.3
```

Best

[pysam_upgrade_failed.txt](https://github.com/pysam-developers/pysam/files/1117873/pysam_upgrade_failed.txt)

### LINKED ISSUE #828: Error with pip install 0.15.2 on OS X
I use `homebrew` with `pip` (not `conda`).
Attempting to install `pysam==0.15.2` on darwin 10.14.5.

Issue
====
`pip install pysam`

Will result in a build failure due to missing `lzma` headers:
```
    htslib/cram/cram_io.c:61:10: fatal error: 'lzma.h' file not found
    #include <lzma.h>
             ^~~~~~~~
    1 error generated.
    error: command 'clang' failed with exit status 1
    ----------------------------------------
```

Solution
======

```
export LDFLAGS=-L/usr/local/opt/xz/lib
export CPPFLAGS="-I/usr/local/opt/xz/include"
```

(or exporting those to wherever your libraries live, that's the default homebrew path) and then running `pip install pysam`.

Not sure where's the best place for this info in your documentation.

jmarshall:
Prevent build failures by ensuring that `configure` uses the same compiler and CFLAGS/etc that will be used later when the Cython and C source itself is compiled.

In particular, the sysconfig $CFLAGS (oddly enough) prevents the compiler from searching /usr/local/include (at least on macOS). Without this PR, configure sees (Homebrew/etc's) /usr/local/include/lzma.h so defines `HAVE_LZMA_H`; later when compiling htslib/cram/cram_io.c that directory is not searched, leading to `<lzma.h> not found` errors. Ensuring that configure also does not see lzma.h there leaves `HAVE_LZMA_H` undefined, which activates HTSlib's os/lzma_stub.h workaround for this header being missing and leads to an lzma-enabled pysam using macOS's system liblzma.dylib.

Fixes #495. Fixes #828.

[It appears I didn't submit this PR back in May, likely because there was no response to my other PRs at the time.]

### DIFF
--- a/devtools/run_tests_travis.sh
+++ b/devtools/run_tests_travis.sh
@@ -43,7 +43,7 @@ conda install -y "samtools=1.9" "bcftools=1.9" "htslib=1.9" xz curl bzip2 conda-
 # Need to make C compiler and linker use the anaconda includes and libraries:
 export PREFIX=~/miniconda3/
 export CFLAGS="-I${PREFIX}/include -L${PREFIX}/lib"
-export HTSLIB_CONFIGURE_OPTIONS="--disable-libcurl --disable-lzma"
+export HTSLIB_CONFIGURE_OPTIONS="--disable-libcurl"
 
 echo "show samtools, htslib, and bcftools versions"
 samtools --version
@@ -54,8 +54,10 @@ bcftools --version
 ~/miniconda3/bin/conda-build devtools/conda-recipe/ --python=$CONDA_PY
 
 # install code from the repository via setup.py
-echo "installing via setup.py from repository"
-python setup.py install
+echo
+echo "============ installing via setup.py from repository ============"
+echo
+python setup.py install || exit
 
 # create auxilliary data
 echo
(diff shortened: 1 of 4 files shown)


AndreasHeger:
Thanks, and apologies for the delays.
Your pull requests are highly appreciated!

PR #833, opened 2019-08-11
Fix lots of broken docs formatting

ExplodingCabbage:
There are currently lots of bits of garbled documentation at https://pysam.readthedocs.io/en/latest/api.html where Napoleon either rendered a multi-line description wrongly, or chucked in some cruft like the text `:param` for no obvious reason. I don't really understand Napoleon (or even vanilla Sphinx/RST), but with some random tinkering around, I've managed to come up with this PR that makes all the really broken stuff I could spot go away.


AndreasHeger:
Thanks!

Issue #799, opened 2019-05-08
VariantFile: [E::vcf_parse_format] - how to handle?


ptrebert:
Hi,
the following is more asking for advice, it may turn into a suggestion for improvement - the basics first:
`Python 3.6 / pysam v0.15.2`
`bcftools / htslib 1.9`

I have the following VCF file:
```
##fileformat=VCFv4.2
##FORMAT=<ID=GT,Number=1,Type=String,Description=Genotype>
##FORMAT=<ID=GQ,Number=1,Type=Float,Description="Genotype Quality">
##contig=<ID=chr1,length=248956422>
#CHROM  POS ID  REF ALT QUAL    FILTER  INFO    FORMAT  NA12878
chr1    13957   .   TC  T   2.57    PASS    .   GT:GQ   1/1:2.57
```
If I understand the VCF 4.2 specification correctly, the `Genotype Quality` field is supposed to be an integer:

> GQ : conditional genotype quality, encoded as a phred quality [...] (Integer)

Nevertheless, if I run `bcftools view <my-vcf.gz>`, I get the full output as above, so the specification does not seem to be enforced (? at least not in the `view` command...). Now I am dealing with a piece of Python/pysam code that uses the header of the above VCF as template for a new/output VCF, and replaces the genotype quality format entry with the (specification-compliant?) integer version (see below) [plus adding new info to the original VariantRecords, not important here]:
```
##fileformat=VCFv4.2
##FORMAT=<ID=GT,Number=1,Type=String,Description=Genotype>
##FORMAT=<ID=GQ,Number=1,Type=Integer,Description="Genotype Quality">
##contig=<ID=chr1,length=248956422>
#CHROM  POS ID  REF ALT QUAL    FILTER  INFO    FORMAT  NA12878
```
With the changed header, `bcftools view <my-vcf.gz>` gives the following error

`[E::vcf_parse_format] Invalid character '.' in 'GQ' FORMAT field at chr1:13957`

Makes sense - now my problem is that the above error seems not to raise an Exception in the Python/pysam code I am dealing with (i.e., at the point when the new VCF is dumped). So I am left with an empty VCF and can only see the `[E::vcf_parse_format]` error message in the log file.

To handle this situation, the following things come to mind:
1. is it possible to get pysam to raise an Exception that could be properly handled in the Python code?
[...]

PR #856, opened 2019-11-04
Attempt to fix of #799.

christianbioinf:
Due to identical return code of `-1` for cases where the end of the file is reached (non-error case) from cases where an error during parsing occurred (e.g. due to incorrect type of field), it is now reading out the error code that may be set inside the `bcf1_t*` record.

In this way, `pysam` can raise an exception and the user is able to properly handle it instead of having a putatively successful read and the burden to check the stderr for any messages like `[E::vcf_parse_format]` or to check whether the total number of records has been processed.


AndreasHeger:
Thanks!

PR #867, opened 2019-11-26
Fix wheel building on TravisCI

### LINKED ISSUE #831: Problem with 0.15.3 wheels on CircleCI
I'm using pysam on CircleCI builds and I've hit an issue with the just-released 0.15.3 binary wheel (pysam-0.15.3-cp36-cp36m-manylinux1_x86_64.whl)

I get the following error:
```
  File "/home/circleci/tskit/python/tests/test_vcf.py", line 40, in <module>
    import pysam
  File "/home/circleci/.local/lib/python3.6/site-packages/pysam/__init__.py", line 5, in <module>
    from pysam.libchtslib import *
  File "pysam/libchtslib.pyx", line 325, in init pysam.libchtslib
AttributeError: type object 'pysam.libchtslib.HTSFile' has no attribute '__reduce_cython__'
```

I'm not sure how to debug the problem, but perhaps a good short-term solution would be to delete the linux wheel from PyPI temporarily so users download and build the source packages as before?

nsoranzo:
- Use cibuildwheel>=1.1.0 to allow specifying the `-L .` option for the auditwheel command. Introduced in joerick/cibuildwheel#211 . Fix #831 .
- Split sdist and linux deploy in 3 separate jobs, otherwise the Travis job times out after 50'.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,11 @@ os:
 
 language: c
 
+stages:
+  - test
+  - name: deploy
+    # if: tag IS present
+
 env:
   matrix:
     - CONDA_PY=2.7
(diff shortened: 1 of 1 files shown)


nsoranzo:
@AndreasHeger Now that binary wheels are successfully tested with `import pysam`, I'd like to re-enable the `twine upload` on tags. If that's OK with you, I can add such a commit to this PR.

AndreasHeger:
Thanks, sorry for the delay.
Yes, let us give it another go.

PR #870, opened 2020-01-06
start htslib1.10

mr-c:


mr-c:
Summary: all tests pass except macOS & CONDA_PY={2.7,3.7} (but macOS + 3.6 is okay!)
The macOS error: 
```
In file included from htslib/hts.c:44:

In file included from /Users/travis/build/pysam-developers/pysam/htslib/cram/cram.h:46:

/Users/travis/build/pysam-developers/pysam/htslib/header.h:227:24: error: must use 'struct' tag to refer to type 'sam_hdr_t'

int sam_hdr_fill_hrecs(sam_hdr_t *bh);

                       ^

                       struct 
```
https://travis-ci.org/pysam-developers/pysam/jobs/633316895#L2870

mr-c:
This PR has been applied to the Debian package of pysam, as we are using htslib 1.10.x now. 

Assistance in fixing the mac OS errors on Python 2.7 & Python 3.7 (so that this PR can be merged) would be appreciated!

jmarshall:
> error: must use 'struct' tag to refer to type 'sam_hdr_t'

This error is what you get when this compilation has picked up a previously installed htslib-1.9 _htslib/sam.h_ rather than the 1.10.x _htslib/sam.h_ bundled within your _pysam_ source tree. This will have come from the `htslib=1.9` conda package installed by _devtools/run_tests_travis.sh_. It isn't immediately clear why this only affects a few platform/conda combinations, but it will come down to the `-I` compiler options added in each case.

Probably trimming and simplifying the `-I` options in play will be the way to fix this.

mr-c:
@jmarshall Thanks for the hint. I don't have access to a macOS system so I can't test it out, though.

If anyone can confirm a fix, then patches or PRs are welcome!

jmarshall:
As mentioned, this is not inherently macOS-specific. Instead it comes from some distutils/setuptools interplay that could in principle be reproduced on Linux.

In any case, I am working up a fix for that. Then an import of upstream 1.10.x should be replayed on top of that fix. Probably by redoing it rather than starting from this PR: samtools & bcftools should be updated too, and this PR has somehow acquired extra copies of all the _htslib/*.h_ headers in other directories etc.

mr-c:
Thanks @jmarshall !

I used the `devtools/import.py` script to do the import of htslib 1.10.2

If this PR is to be replaced, you'll likely still need https://github.com/pysam-developers/pysam/pull/870/commits/790fd114aa8d27c795f94d6b4ddfb9331217b5bd

mr-c:
@jmarshall A previous version of this PR also updated samtools/bcftools, but then a bunch of tests broke due to some extra output in v1.10.x, so I rolled it back to this version

AndreasHeger:
Thanks, @mr-c for making a start on this, and apologies for being slow out of the block. Thanks @jmarshall  for your assistance. I can give it a go

jmarshall:
Superseded by PR #905.

PR #874, opened 2020-01-14
Fix a broken import

ExplodingCabbage:
Without this change, installing from the repo (with `pip3 install .` or `python3 setup.py install`) produces a broken installation that fails on this import when I try to do `import pysam`.

(I'm confused about why the version I get by running `pip3 install pysam` doesn't have the same issue. It seems to already have this change. Is there something in the build pipeline that rewrites this import, that's missing when I install locally? Or, @AndreasHeger, did you make this change locally before pushing to PyPI and not commit it? Sorry if this is a dumb question; I'm inexpert at how setuptools works.)


AndreasHeger:
Hi @ExplodingCabbage , (Pratchett fan?) the issue could be that the following usually does not work:

ExplodingCabbage:
Nope, this PR is not motivated by that mistake (although I [have previously made it](https://github.com/pysam-developers/pysam/issues/860#issuecomment-572046125) so your suspicion was reasonable).

If I install pysam from source on the current master branch and then try to import it from somewhere with no `pysam` directory in the cwd, I get this:

```
mark@fractal:/$ python3
Python 3.8.0 (default, Oct 28 2019, 16:14:01) 
[GCC 9.2.1 20191008] on linux
>>> import pysam
  File "/usr/local/lib/python3.8/dist-packages/pysam/__init__.py", line 30, in <module>
    from pysam.samtools import *
  File "/usr/local/lib/python3.8/dist-packages/pysam/samtools.py", line 1, in <module>
    from utils import PysamDispatcher
ModuleNotFoundError: No module named 'utils'
```

This seems to be the only import that needs changing to let me do `import pysam` successfully and I am confused about why this one import is written differently to all others and why you (I guess?) don't have the same problem.

> (Pratchett fan?)

Not since I was a teen, and I didn't read enough to know what you're referencing, but perhaps now I will. :)

AndreasHeger:
Thanks, I will try changing this to pysam.utils.

In one of Pratchett's book is an Exploding Cabbage Stamp.

AndreasHeger:
The import has changed via a separate PR, I am closing this.
Thanks!

PR #875, opened 2020-01-14
Bump minimum Cython version

ExplodingCabbage:
This is the version @pm67nz indicates is needed for creating a build from source in https://github.com/pysam-developers/pysam/issues/860#issuecomment-572451657, so I figure we might as well enforce it everywhere?


AndreasHeger:
Yes, many thanks.

PR #878, opened 2020-01-15
Restore twine upload on tag

nsoranzo:
Follow-up on #867 .


nsoranzo:
The upload of wheels worked fine and they seem to work correctly, at least in a simple `import pysam` test.

Issue #39, opened 2014-09-11
BAM file created by BWA with -R argument causes 'ValueError: unknown field code 'SM' in record 'PG''


pmarks:
The BAM file was created with a BWA 0.7.9a-r786.
It inserted the following @PG header line:

```
@PG     ID:bwa  PN:bwa  VN:0.7.9a-r786  CL:bwa mem -p -t 8 -M -R @RG    ID:None SM:None /mnt/data/hg19.fa /mnt/analysis/default-0.fastq
```

The -R argument to bwa is used to supply the read group header, and BWA help instructs the user to supply tab-delimted fields to that argument:

```
       -R STR     read group header line such as '@RG\tID:foo\tSM:bar' [null]
```

The problem is that BWA puts the whole command-line, including the tab-delimited -R argument in the `@PG` `CL:` field.  But the SAM header format apparently doesn't have a way to escape tabs in the values, so pysam is interpreting the SM:None field as a field in the `@PG` header, rather than as part of the `CL:` string.  Seems that the real problem is in the SAM spec. Should pysam be made more forgiving of this case?

```
In [1]: import pysam

In [2]: sf =pysam.Samfile("/mnt/data/output.bam")

In [3]: sf.header
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-40307eadca76> in <module>()
----> 1 sf.header

/mnt/home/pat/code/pl2/devpipes/anaconda/1.9.2/lib/python2.7/site-packages/pysam/csamfile.so in pysam.csamfile.Samfile.header.__get__ (pysam/csamfile.c:12899)()

ValueError: unknown field code 'SM' in record 'PG'

In [4]: pysam.version.__version__
Out[4]: '0.8.0'
```

smowton:
Ran into a similar problem, created https://github.com/lh3/bwa/pull/46 to address this in BWA.

PR #890, opened 2020-02-05
Update test suite to work with samtools/etc 1.10.x external commands

jmarshall:
The legacy import command was broken in Samtools 1.10 (hat tip @mr-c and see also samtools/samtools#1183) and has been removed from future Samtools (see samtools/samtools#1185).

Separately from and prior to any updating of the bundled htslib/samtools/etc to 1.10 or later (this fails even now with the bundled samtools 1.9, as an external `samtools` is used which may be 1.10), the test suite needs to stop using this legacy command. Use the equivalent `samtools view` command instead.


jmarshall:
Also most of these almost certainly don't need the antiquated `-t ex1.fa.fai` option, but it doesn't really matter and this PR makes the minimal change and leaves them as is.

unode:
The additional strictness of `samtools` `1.10` when parsing the header is also making tests fail on lines starting with `@x1`, `@x2` and `@x3`.

This happens with:
https://github.com/pysam-developers/pysam/blob/c818db502b8f8334e7bf29060685114dd9af9530/tests/pysam_data/example_user_header.sam#L4

unode:
To be clear, the SAM specification states:

> ... `TAG` is a two-character string that defines the format
> and content of `VALUE`. Thus header lines match 
> `/^@(HD|SQ|RG|PG)(\t[A-Za-z][A-Za-z0-9]:[ -~]+)+$/`or`/^@CO\t.*/`

and therefore `@x...` tags are rejected. `X...` tags can still exist but only in alignment fields.

I'm not sure if this defeats the purpose of the test but converting them to comments (`@CO`) makes `samtools` happy.

jmarshall:
You are correct — that's also a very unfortunate test case. I'll add a fix to this PR.

unode:
Thanks!

unode:
Seems like it doesn't end there:
```
samtools view -bS rg_with_tab.sam > rg_with_tab.bam
[E::sam_hrecs_error] Malformed key:value pair at line 3: "@PG   ID:bwa  PN:bwa  VN:0.7.9a-r786  CL:bwa mem -p -t 8 -M -R @RG    ID:None SM:None /mnt/data/hg19.fa       /mnt/analysis/default-0.fastq"
samtools view: failed to add PG line to the header
```
:cry:

jkbonfield:
Indeed, that is also an illegal SAM header.

The regexp listed in https://github.com/samtools/hts-specs/blob/master/SAMv1.pdf shows that you can only have tab in your header fields if they're in CO lines.  It doesn't elaborate on how to work around this though.  I guess "\t" or something is appropriate.

jmarshall:
I've changed or removed the test cases that were failing due to Samtools 1.10's increased header parsing and hence requirement for headers to be valid, as described in the commit message.

I've also worked around Samtools 1.10's inconvenient behaviour change whereby it now adds `@PG` headers by default. This can be suppressed with `--no-PG` but that option is a syntax error with earlier versions of samtools, which pysam still needs to work with. (Compare the related samtools/samtools#1175.) Fortunately the problematic _Makefile_ already uses GNU Make constructs, so we can use a `$(shell)` kludge to probe for the existence of the option.

Finally _ex3.sam_ contained records with incorrect NM values. Samtools 1.10 preserves these incorrect values when converting to CRAM, while previous versions effectively replaced them with the correct value. This was leading to another 8 test failures with samtools 1.10. We're testing pysam not samtools's format conversion, so I've corrected the values in the input data file and adjusted the SAM/BAM tests accordingly.

This fixes all test failures for me locally; we will see more with <https://travis-ci.org/jmarshall/pysam/builds/653253555>.

unode:
> I guess "\t" or something is appropriate

`\t` is typically used to represent a TAB character. Couldn't tell from your phrasing if this is what you meant.

jmarshall:
Re tab characters, that test case reflects a legacy pysam workaround (#39) for buggy BWA behaviour in simple uses of BWA's monolithic `-R` option. There was a suggestion for BWA to avoid outputting invalid headers by encoding the string for `@PG` with two characters `'\\', 't'` itself, but sadly lh3/bwa#46 was never merged. BWA's answer was that it decodes `\t` itself, so as long as you can quote your backslashes enough that BWA **itself** gets to do the expanding, it will be printed as a tab in the `@RG` header but `\t` in the `@PG` header. Unfortunately [this is in general infeasible](https://sourceforge.net/p/bio-bwa/mailman/message/26744681/), and I was [never able to convince the author](https://sourceforge.net/p/bio-bwa/mailman/message/26729447/) to sidestep the entire problem by using separate options like `samtools addreplacerg -r ID:foo -r SM:bar` does.

But I digress. As noted in the commit message, pysam's PG-CL-with-tabs workaround is no longer feasible to test except in the original SAM text and the workaround is somewhat misguided anyway, so I have just defanged the test case by changing them to spaces (as sufficiently quoting the expected output in _AlignmentFile_test.py_ would be tedious if they were changed to `\t` — yes, I know about Python raw strings). In future, it might be worth rethinking this pysam extension and removing it and this now badly-named test case.

jmarshall:
@AndreasHeger: This PR now passes all tests when used together with external samtools/bcftools/etc provided by [samtools/etc 1.9](https://travis-ci.org/pysam-developers/pysam/builds/653424106) or by [samtools/etc 1.10](https://travis-ci.org/jmarshall/pysam/builds/653424636). I hate to suggest rolling yet another htslib-1.9-using pysam 0.15.x point release so soon after sorting out the Cython/Py3.8 mess, but it might be a good idea…

PR #910, opened 2020-04-15
Fix incorrect quoting in VariantFile contig records

### LINKED ISSUE #909: VariantHeaderContigs: contigs added with a length attribute are silently ignored
When I use pysam to create a new VariantFile from scratch, I cannot add contig meta-information to the header if that meta-information contains a length attribute. The contig is not added to the header and no error is reported:

```python
>>> header = pysam.VariantHeader()
>>> header.contigs.add("1", length=249250621)
>>> header.contigs.items()
[]
```

The source of this issue seems to be that htslib requires the length attribute to be an unquoted string value, whereas pysam automatically quotes the values of all but a selected few attributes. Right now, the decision which attribute values to quote is made in the `add_meta` method of the `VariantHeader` class. The following code fragment from that method shows that all attribute values are quoted unless the attribute name is one of ID, Number, or Type.

```python
quoted = True
if key in set(("ID", "Number", "Type")):
    quoted = False
```

Adding `"length"` to the above set fixes the reported issue. However, similar issue may arise with other types of meta-information. Besides, attribute names and their quoting requirements may depend on the specific type of meta-information. For that reason, I propose a more generic fix for this issue. In my proposed solution, the quoting format is passed to the ```add_meta``` method by upstream functions that know which attributes should not be quoted. For example, the ```VariantHeaderContigs.add``` method would look like this:

```python
def add(self, id, length=None, **kwargs):
    """Add a new contig record"""
    if id in self:
        raise ValueError('Header already exists for contig {}'.format(id))

    items = [('ID', unquoted_str(id))]
    if length is not None:
       items.append(("length", unquoted_str(length)))
    items += kwargs.items()
    self.header.add_meta('contig', items=items)
```

The call to ```unquoted_str``` tags the value as an unquoted string value. The ```add_meta``` method then uses this to pass the quoting format to htslib using the following check, which replaces the comparison to the set of hard-coded attribute names.

```python
quoted = not isinstance(value, unquoted_str)
```

Let me know if you would consider integrating this solution. If so, I will make a pull request.

scanisius:
Fixes #909. This pull request addresses the quoting issue when adding contigs to a VariantFile's header. In addition, I implemented a more generic solution that disables quoting of attribute values by tagging those values with `unquoted_str`. This may be helpful for addressing similar issues with other types of meta-information.

### DIFF
--- a/pysam/libcbcf.pyx
+++ b/pysam/libcbcf.pyx
@@ -100,6 +100,8 @@ from cpython.version cimport PY_MAJOR_VERSION
 
 from pysam.libchtslib cimport HTSFile, hisremote
 
+from pysam.utils import unquoted_str
+
 
 __all__ = ['VariantFile',
            'VariantHeader',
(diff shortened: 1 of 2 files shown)


AndreasHeger:
Thanks, @scanisius ,@bioinformed , this looks good to me.

Issue #769, opened 2019-03-05
Update docs


kyleabeauchamp:
```

Author: | Andreas Heger, Kevin Jacobs and contributors
-- | --
Jul 27, 2018
0.15.0

The current version wraps htslib-1.7, samtools-1.7 and bcftools-1.6.



```

jmarshall:
Since 84613844b82add695a4e55322c0d0a6ed3f34498 _devtools/import.py_ automatically updates these version numbers in the documentation.

PR #925, opened 2020-05-31
{AH} bump version to 0.16.0

AndreasHeger:
Release candidate for v0.16.0 release - apologies for the delay.

Hi @bioinformed  and @jmarshall , is there anything you think I should wait for?

Otherwise I will update the docs and push out the release.

### DIFF
--- a/doc/release.rst
+++ b/doc/release.rst
@@ -6,11 +6,24 @@ Release 0.16.0
 ==============
 
 This release wraps htslib/bcftools version 1.10.2 and samtools version
-1.10. Additional bugfixes:
+1.10. The following bugs reported against pysam are fixed due to this:
 
-* [#909] Fix incorrect quoting in VariantFile contig records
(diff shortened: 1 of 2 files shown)


jmarshall:
I added some more release notes entries — please edit as you see fit. I haven't done a full check that the list is now complete; in particular, there's a couple of others (e.g. #769) that are fixed too but uninteresting for the release notes.

There's a couple of issues I'm looking at as free time permits, but nothing that should be waited for. The next samtools/etc release is on the horizon but probably still measured in weeks, so shouldn't be waited for. I have some fixes for the bioconda pysam package that I must finish up testing, and if the timing was right that could be deployed at the same time as a new pysam release.

kevinjacobs-progenity:
I've run some local tests and don't see any issues.  Thanks, @jmarshall!

AndreasHegerGenomics:
Many thanks for this, @jmarshall , I will release tonight.

PR #941, opened 2020-06-30
Fixed "idx_test_and_fetch" to return proper index filename string length

jvierstra:
This change makes the hts.c consistent with the official htslib codebase and fixes a bug where the locally downloaded index could not be found when accessing remote tabix files.

### DIFF
--- a/htslib/hts.c
+++ b/htslib/hts.c
@@ -3395,6 +3395,7 @@ static int idx_test_and_fetch(const char *fn, const char **local_fn, int *local_
             fclose(local_fp);
             free(s.s);
             *local_fn = p;
+            *local_len = e-p;
             return 0;
         }
 


jmarshall:
Essentially you want to backport part of samtools/htslib#1014, which landed in htslib after the 1.10.2 release.

This will be picked up in due course when pysam is updated after the next htslib/etc release, which is expected to be a matter of weeks from now. Is this something that is causing you significant difficulty and you would hope to have fixed sooner than that?

jvierstra:
I fixed it on my fork -- no time pressure.  It's just an annoying bug that causes the filenames of the bam/tabix index files to get mangled when accessing remote files repeatedly.

jmarshall:
Thanks for proposing this PR. The fix has now been added to pysam by PR #962, which updated the bundled htslib/etc to 1.11 — which, in turn, include this fix.

(Now we just have to get the wheels cranking on a release containing the fix.)

Issue #865, opened 2019-11-21
adding a new line character at the example of pysam.FastxFile to write a fastx file


edwwlui:
at `pysam/pysam/libcfaidx.pyx`, the example of pysam.FastxFile has to add a new line character so as to write a fastx file in the valid format.

`cdef class FastxFile:`

...

    Examples
    --------
    >>> with pysam.FastxFile(filename) as fh:
    ...    for entry in fh:
    ...        print(entry.name)
    ...        print(entry.sequence)
    ...        print(entry.comment)
    ...        print(entry.quality)
    >>> with pysam.FastxFile(filename) as fin, open(out_filename, mode='w') as fout:
    ...    for entry in fin:
    ...        fout.write(str(entry)+"\n")
    """

PR #960, opened 2020-10-12
adding a new line character at the example of pysam.FastxFile to write a valid fastx file

edwwlui:
otherwise quality line would be mixed up with the header line


jmarshall:
Thanks, I've updated this to use the same `+ '\n'` as in your issue #865 and merged it as daee1a9837a91c2909dde095cb602dcf1538a612.

PR #963, opened 2020-10-30
Expose VCF classes

nh13:
This is extremely useful and important when we want to use the typing library.  For example, I have a lot of methods that pass around a `VariantRecordSample` but I am unable to annotate the type since the class is not able to be referenced.


nh13:
@jmarshall or @AndreasHeger I'd appreciate a review, or a pointer to a maintainer to review.  I am open for reducing the # of exposed classes.

kevinjacobs-progenity:
I'm the original developer and fully support this change.  I've wanted to add better class documentation and static typing, but spare time is just so limited.  Thanks for your contribution and I'll commit to reviewing any future PRs.

kevinjacobs-progenity (approved):

nh13:
Thank you! I’ll make follow on PRs once I start using the types of I see anything. Thank you for the Friday PR review!

PR #962, opened 2020-10-20
Update imports to htslib/samtools/bcftools 1.11

jmarshall:
Import htslib, samtools, and bcftools 1.11:

* First alter _tests/pysam_data/ex1.sam.gz_ (and files built from it) to avoid exercising a htslib-1.11 bug as described in https://github.com/samtools/htslib/pull/1031#issuecomment-697377039; here we are testing the pysam bindings not this particular behaviour, so simply avoiding the problem is legitimate.

   (You will need to regenerate these files with `git clean -xf tests/pysam_data; make -C tests/pysam_data` or similar before running the pysam tests with htslib-1.11.)

* Import the 1.11 releases, and add the new `ampliconclip` and `ampliconstats` subcommands to _pysam/samtools.py_.

* Backport samtools/htslib#1150 and samtools/htslib#1184 to avoid potential plugin problems.

* Bump version numbers in documentation.

PR #977, opened 2020-12-18
docs: fix simple typo, requring -> requiring

timgates42:
There is a small typo in htslib/cram/mFILE.c.

Should read `requiring` rather than `requring`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md


jmarshall:
Code in _htslib/…_ is imported from [htslib](https://github.com/samtools/htslib), so it's better not to cause the copy here to diverge. This particular typo has already been fixed upstream and will be corrected when PR #962 is merged.

It would be worth correcting spelling in the top-level files and _pysam/…_, but please batch any corrections up rather than presenting them as individual PRs.

PR #978, opened 2020-12-31
Finishes commenting libctabixproxies

homonecloco:
I am having issues when running under Python-3.9 installed on MacOS Big Sur using homebrew.  I believe this is related to the issue  pysam-developers/pysam#443 . I patched this on my local computer by removing the mentions to libctabixproxies in the __init__.py that remain.

The error I was getting was the following: 

```
  File "/usr/local/lib/python3.9/site-packages/pysam/__init__.py", line 43, in <module>
    libctabixproxies.__all__ +\
NameError: name 'libctabixproxies' is not defined
```

### DIFF
--- a/pysam/__init__.py
+++ b/pysam/__init__.py
@@ -40,7 +40,6 @@ __all__ = \
     libcbcf.__all__ +\
     libcbgzf.__all__ +\
     libcfaidx.__all__ +\
-    libctabixproxies.__all__ +\
     libcalignmentfile.__all__ +\
     libcalignedsegment.__all__ +\
     libcsamfile.__all__ +\
@@ -87,7 +86,7 @@ def get_libraries():
     # Note that this list does not include libcsamtools.so as there are
     # numerous name conflicts with libchtslib.so.
     dirname = os.path.abspath(os.path.join(os.path.dirname(__file__)))
-    pysam_libs = ['libctabixproxies',
+    pysam_libs = [
                   'libcfaidx',
                   'libcsamfile',
                   'libcvcf',


jmarshall:
I can't reproduce this problem (at least, not using a development build). How do you have pysam installed — via a wheel or other package perhaps?

In any case, there were three commits during #443 that affected this:

1. 44ebba4c38b4a69eff8293181673c37a9fff008d added the imports and added `libctabixproxies.__all__` to `__all__`
2. 6adb0cda135642d4dae8c816f8b370850e724dd4 commented out the imports
3. ba80ddd9f682df409e4393e085d34317d2e5dce0 added `__all__` to _libctabixproxies.pyx_, which did not previously define `libctabixproxies.__all__`

Thus the combination of _\_\_init\_\_.py_ imports and `__all__` defined has not been tried.

So I suspect that, now that _libctabixproxies.pyx_ defines `__all__`, instead of removing the remaining references, this could also be fixed by uncommenting the imports. Thus:

```diff
--- a/pysam/__init__.py
+++ b/pysam/__init__.py
@@ -11,8 +11,8 @@ import pysam.libcfaidx as libcfaidx
 from pysam.libcfaidx import *
 import pysam.libctabix as libctabix
 from pysam.libctabix import *
-# import pysam.libctabixproxies as libctabixproxies
-# from pysam.libctabixproxies import *
+import pysam.libctabixproxies as libctabixproxies
+from pysam.libctabixproxies import *
 import pysam.libcsamfile as libcsamfile
 from pysam.libcsamfile import *
 import pysam.libcalignmentfile as libcalignmentfile
```

Could you try with that change instead of removing the `libctabixproxies.__all__ +\` line? Hopefully that will allow `import pysam` without any errors too.

PR #981, opened 2021-01-12
Add support for pathlib.Path as filename arguments 

arogozhnikov:
Context: 
- python 3 has two ways to pass paths: str/bytes and pathlib.Path (very convenient BTW)
- most functions and libs (e.g. open, pandas) handle pathlib.Path very well, and can use for filenames instead of `str`s. However that doesn't work for e.g. pysam.AlignmentFile and I keep forgetting this. This small change fixes the problem.

### DIFF
--- a/pysam/libcutils.pyx
+++ b/pysam/libcutils.pyx
@@ -6,7 +6,11 @@ import tempfile
 import os
 import io
 from contextlib import contextmanager
-from pathlib import Path
+try:
+    from pathlib import Path
+except:
+    # for Python 2
+    Path = str
 
 from cpython.version cimport PY_MAJOR_VERSION, PY_MINOR_VERSION
 from cpython cimport PyBytes_Check, PyUnicode_Check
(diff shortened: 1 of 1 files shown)


arogozhnikov:
CI passed 
https://travis-ci.org/github/pysam-developers/pysam/builds/754082159

arogozhnikov:
ping

jmarshall:
Have you tried doing this without your patch? I believe the `os.fsencode` call a few lines below your suggested patch already has the same effect (since PR #392), and e.g. `fn = Path("foo.bam"); f = pysam.AlignmentFile(fn, "r")` already works for me.

arogozhnikov:
@jmarshall very good point. I didn't notice fsencode supports that.

For some reason that did not work for me when I wrote this patch, but now it works as it should.
Could be that I used python 3.5 where fsencode did not support Path objects? Unlikely, all my code has f-strings.

I'll have a bit more testing, will close PR if issue won't appear again.

arogozhnikov:
seems problem is solved, thanks again for pointing @jmarshall

Issue #998, opened 2021-03-20
Failed reading ASCII encoded vcf files with special characters


olingerc:
First, thanks for the great library!

I have a ASCII encoded VCF file with non-standard characters in the INFO field. A minimal example would be:
```
##fileformat=VCFv4.2
##contig=<ID=1,length=249250621>
##INFO=<ID=ILoveFrance,Number=R,Type=String,Description="Everybody should">
##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	Sample
1	65529	.	C	T	16.65	.	ILoveFrance=Télomère	GT	0/1
```
In windows simply use notepad and make sure that when saving you use "ANSI" encoding.

The following code will fail with `UnicodeDecodeError` with pysam version 0.16.0.1
```python
from pysam import VariantFile

vcf_in = VariantFile(filename="test.vcf", mode="r")

for v in vcf_in.fetch():
    print(v.info.items())
```

This line in the pysam code raises the exception in [libcbcf.py](https://github.com/pysam-developers/pysam/blob/746e4d9bd149722f2dbdb99b1e15760a7c2b7979/pysam/libcbcf.pyx#L287):

```python
value = tuple(v.decode('utf-8') if v and v != bcf_str_missing else None for v in b.split(b','))
```

Does that mean that pysam will always force to UTF-8? I can of course convert the file to UTF-8 beforehand, but my tool needs to be able to work with every possible incoming VCF file. Is there any way to handle the exception and try to decode the string myself?

Full exception:

```
Traceback (most recent call last):
  File "test.py", line 6, in <module>
    print(v.info.items())
  File "pysam/libcbcf.pyx", line 2750, in pysam.libcbcf.VariantRecordInfo.items
  File "pysam/libcbcf.pyx", line 2741, in iteritems
  File "pysam/libcbcf.pyx", line 555, in pysam.libcbcf.bcf_info_get_value
  File "pysam/libcbcf.pyx", line 287, in genexpr
  File "pysam/libcbcf.pyx", line 287, in genexpr
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 0: invalid continuation byte
```

bioinformed:
Quick answer:

   - Your input file isn't valid VCF, so you'll need to encode it as UTF-8
   before parsing it with pysam.

Long answer:

[...]

PR #1001, opened 2021-03-25
Add get/set_encoding_error_handler() to control UTF-8 conversion

jmarshall:
Python has impressively flexible facilities to control how the [decode() function](https://docs.python.org/3/library/stdtypes.html#bytes.decode) deals with unexpected input.

The easiest way to enable scripts to control pysam's use of the `decode()` function is to expose this as a global variable via getter/setter methods, similarly to the existing `get_verbosity()`/`set_verbosity()`.

For convenience, also define an error handler that simply interprets invalid/unexpected UTF-8 bytes as ISO-8859-1 (Latin-1) characters. Make it available via `set_encoding_error_handler('pysam.latin1replace')`.

Issue #998 is about pysam's basic UTF-8 parsing refusing to parse fields containing Latin-1-encoded accented letters. This PR would allow scripts to parse such fields by adding:

```python
pysam.set_encoding_error_handler('pysam.latin1replace')
```

before calling functions that decode such fields. Other built-in standard error handlers such as `'replace'` and `'backslashreplace'` may also be of interest.

Parsing as UTF-8 and interpreting invalid input as Latin-1 is problematic because two or more adjacent Latin-1 characters can be misinterpreted as a validly UTF-8-encoded unrelated character. (Probably no such Latin-1 strings are intelligible in French or any other language.) Hence it may be better to also expose the encoding used via global getter/setters so that scripts could instead write

```python
pysam.set_encoding('latin-1')
```

to specify that their VCF v4.2 files should NOT be interpreted as UTF-8 but purely as Latin-1.

I have not added such functions because `force_str()` et al's optional arguments are currently buggy — they supply the original value of `TEXT_ENCODING` (UTF-8), not the current value, so there's little point in adding functions to change the value of `TEXT_ENCODING`. But I would be happy to rework the optional `encoding=` arguments and add such a getter/setter (probably IMHO in addition to the error handler one) if other maintainers prefer.

### DIFF
--- a/pysam/libcutils.pyx
+++ b/pysam/libcutils.pyx
@@ -6,6 +6,7 @@ import tempfile
 import os
 import io
 from contextlib import contextmanager
+from codecs import register_error
 
 from cpython.version cimport PY_MAJOR_VERSION, PY_MINOR_VERSION
 from cpython cimport PyBytes_Check, PyUnicode_Check
@@ -89,6 +90,8 @@ cpdef qualities_to_qualitystring(qualities, int offset=33):
 def latin1_replace(exception):
     return (chr(exception.object[exception.start]), exception.end)
 
+register_error('pysam.latin1replace', latin1_replace)
+
 
 cdef str ERROR_HANDLER = 'strict'
 
@@ -98,10 +101,6 @@ cpdef get_encoding_error_handler():
 cpdef set_encoding_error_handler(name):
     global ERROR_HANDLER
     previous = ERROR_HANDLER
-    if name.startswith('pysam.'):
-        from codecs import register_error
-        register_error('pysam.latin1replace', latin1_replace)
-
     ERROR_HANDLER = name
     return previous
 
@@ -137,7 +136,7 @@ cdef bytes encode_filename(object filename):
         raise TypeError("Argument must be string or unicode.")
 
 
-cdef bytes force_bytes(object s, encoding=TEXT_ENCODING):
+cdef bytes force_bytes(object s, encoding=None, errors=None):
     """convert string or unicode object to bytes, assuming
     utf8 encoding.
     """
@@ -146,37 +145,37 @@ cdef bytes force_bytes(object s, encoding=TEXT_ENCODING):
     elif PyBytes_Check(s):
         return s
     elif PyUnicode_Check(s):
-        return s.encode(encoding)
+        return s.encode(encoding or TEXT_ENCODING, errors or ERROR_HANDLER)
     else:
         raise TypeError("Argument must be string, bytes or unicode.")
 
 
-cdef charptr_to_str(const char* s, encoding=TEXT_ENCODING):
+cdef charptr_to_str(const char* s, encoding=None, errors=None):
     if s == NULL:
         return None
     if PY_MAJOR_VERSION < 3:
         return s
     else:
-        return s.decode(encoding, ERROR_HANDLER)
+        return s.decode(encoding or TEXT_ENCODING, errors or ERROR_HANDLER)
 
 
-cdef charptr_to_str_w_len(const char* s, size_t n, encoding=TEXT_ENCODING):
+cdef charptr_to_str_w_len(const char* s, size_t n, encoding=None, errors=None):
     if s == NULL:
         return None
     if PY_MAJOR_VERSION < 3:
         return s[:n]
     else:
-        return s[:n].decode(encoding, ERROR_HANDLER)
+        return s[:n].decode(encoding or TEXT_ENCODING, errors or ERROR_HANDLER)
 
 
-cdef bytes charptr_to_bytes(const char* s, encoding=TEXT_ENCODING):
+cdef bytes charptr_to_bytes(const char* s, encoding=None, errors=None):
     if s == NULL:
         return None
     else:
         return s
 
 
-cdef force_str(object s, encoding=TEXT_ENCODING):
+cdef force_str(object s, encoding=None, errors=None):
     """Return s converted to str type of current Python
     (bytes in Py2, unicode in Py3)"""
     if s is None:
@@ -184,10 +183,19 @@ cdef force_str(object s, encoding=TEXT_ENCODING):
     if PY_MAJOR_VERSION < 3:
         return s
     elif PyBytes_Check(s):
-        return s.decode(encoding, ERROR_HANDLER)
+        return s.decode(encoding or TEXT_ENCODING, errors or ERROR_HANDLER)
     else:
         # assume unicode
         return s
+
+
+cdef decode_bytes(bytes s, encoding=None, errors=None):
+    """Return s converted to current Python's str type,
+    always decoding even in Python 2"""
+    if s is None:
+        return None
+    else:
+        return s.decode(encoding or TEXT_ENCODING, errors or ERROR_HANDLER)
 
 
 cpdef parse_region(contig=None,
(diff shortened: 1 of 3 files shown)


jmarshall:
Doing this via a global and getter/setter is a bit limiting.

For example, some might want to open two files simultaneously: a VCFv4.3 file to be read as UTF-8, and a CLC Workbench-written VCFv4.2 file to be read as Latin-1. So there would be some advantage in making this a per-file setting, and propagating the setting down through all the subroutines so it can be specified in each `force_str`/etc call… Not Fun. (But it might not be too bad if most of the functions that call `force_str`/etc are member methods of `VariantFile` and other file classes.)

Implementing it via a single pysam global is a lot simpler. Hopefully it provides 90% of the needed functionality…

jmarshall:
Renamed the functions to `get/set_encoding_error_handler()` as that's clearer and follows the precedent of some existing `get/set_encoding()` functions.

pysam/libcutils.pyx
```diff
@@ -144,7 +166,7 @@ cdef charptr_to_str_w_len(const char* s, size_t n, encoding=TEXT_ENCODING):
     if PY_MAJOR_VERSION < 3:
         return s[:n]
     else:
-        return s[:n].decode(encoding)
+        return s[:n].decode(encoding, ERROR_HANDLER)
 
 
 cdef bytes charptr_to_bytes(const char* s, encoding=TEXT_ENCODING):
```
kevinjacobs-progenity, line 172: @jmarshall : Thanks again for tackling this.

Replacing the default with `None` and adding:
```
if encoding is None:
    encoding = TEXT_ENCODING
```
will allow the default to track the global variable.  This isn't so important for encoding, but the same technique will be needed to support per-file error handlers.  After this PR lands, I'd like to try to add the the error handler as an attribute to the base HTSFile class and see how hard it is to get true per-file settings.  Even if users only use the global setting API, this will also allow the error handler to be fixed at the time a file is opened, rather than changing unexpectedly as a side-effect throughout its lifetime.

jmarshall:
My thinking on the eventual flexible API for all this is that file classes like `VariantFile` (or the base `HTSFile`, good point @kevinjacobs-progenity) should have optional constructor arguments and/or getter/setters:

```python
encoding=None
encoding_error_handler=None
```

that, if set, would override the Pysam default set by

```python
pysam.set_encoding(…)
pysam.set_encoding_error_handler(…)
```

and if you didn't use those either, the default Pysam defaults would be `utf-8` and `strict`.

pysam/libcutils.pyx
jmarshall: Yes indeed; an earlier version of this added an `errors=None` as an optional argument to `force_str`/etc as well. I guess the futureproofing thing to do would be to change the signatures to e.g.

```python
def charptr_to_str_w_len(const char* s, size_t n, encoding=None, errors=None):
```
and do that fixing up inside these three or four _libcutils.pyx_ functions as you suggest.

Good point about the constructor needing to give the ability to fix these fields so that file is unaffected by later changes to the global setting.

jmarshall:
Do we like the `encoding_error_handler` terminology btw? I guess `encoding_errors`/ `get_encoding_errors`/ `set_encoding_errors` would be closer to the underlying Python `encodings` + `errors` arguments to `decode()`, but outside of that context I think making it clear that this is a handler (not e.g. a list of the actual errors encountered or something) is useful.

kevinjacobs-progenity:
I am happy with `encoding_error_handler`, since pysam's API is complex and has trended toward using longer and more explicit attribute and argument names.

pysam/libcutils.pyx
```diff
@@ -82,7 +82,29 @@ cpdef qualities_to_qualitystring(qualities, int offset=33):
...
+
+cdef str ERROR_HANDLER = 'strict'
+
+cpdef get_encoding_error_handler():
+    return ERROR_HANDLER
+
+cpdef set_encoding_error_handler(name):
+    global ERROR_HANDLER
+    previous = ERROR_HANDLER
+    if name.startswith('pysam.'):
```
kevinjacobs-progenity, line 101: Is this check intended to be more general?  It isn't obvious why you're only checking the prefix.  (Feel free to tell me to hold my horses until you're ready for it to be reviewed.)

jmarshall: The user will be calling `set_encoding_error_handler()` with either a standard handler, one they wrote themselves, or one that pysam conveniently supplies. We only need to register the pysam-supplied ones if the user is actually asking for a pysam-supplied one, and in the interests of “libraries shouldn't do weird stuff if they don't need to” this check avoids registering ours if the user is using a standard one or their own.

The check could be `name == 'pysam.latin1replace'` but I'm imagining the list might expand in future. In particular, I might add a `pysam.latin15replace` as well…

That was the idea anyway… I suppose once we add file class `set_encoding_error_handler()` methods as well, it would make more sense to move this out of the global `set_encoding_error_handler` function and just always register them in whatever the appropriate form of `__init__` processing is.

kevinjacobs-progenity: That makes perfect sense now that I understand where the concept is/was/may still be going.

jmarshall:
Thanks for reviewing, @kevinjacobs-progenity. I've finally got back to this, and I think made all the changes we discussed:

* Reworked the optional argument handling so as to be ready for settings coming from the HTSFile in future;

* Turns out this was failing on Python 2.7 because reusing `force_str()` wasn't really the right thing to do. Instead I've added another function that always decodes UTF-8, even on Python 2 — which is appropriate for these VCF fields.

pysam/libcutils.pyx
jmarshall: I've updated this so all the relevant functions have `encoding=None, errors=None` optional arguments.

pysam/libcutils.pyx
jmarshall: I've changed it so that it always just registers the handler name when _libcutils.pyx_ is loaded. (I'm no expert, but it seems like just putting the code at the top-level of the _.pyx_ file so it is executed on load is “the right form of `__init__` processing” in this context…?)

This paves the way for future file class `set_encoding_error_handler()` methods (e.g. for HTSFile) — if user code was using those, then the global `set_encoding_error_handler()` function would likely never be called (so it would be a bad place to do this init processing!).

kevinjacobs-progenity:
@jmarshall : Many thanks!  I'll review again over the weekend. 

On a related note, I've had several user requests to add automatic URL encoding and decoding to escape characters in INFO and GT string values.  It seems like it would be relatively easy to do this, building on this work.  Does this make sense to do here or should we wait until htslib adds this functionality?

kevinjacobs-progenity:
Arg.  I completely forgot I _had_ a working branch for HTSlib with percent escaping of string values.  I was changing jobs and I must be getting a bit senile.  I'll see how much of that original work is still relevant and decide how to proceed.

olingerc:
Sorry to bump this again. Do you have any news about this? I also haven't found any new release since 0.16.0.1. Is master safe to use?

kevinjacobs-progenity:
The main branch should be safe to use.  And I've lightly tested this branch.  However, life is currently complicated and I have not made substantive progress on implementing finer-grained control of encoding.  I'll set aside time in the next few days to give this PR a more thorough review so that it can be merged.

jmarshall:
I think this branch is itself feature-complete and is a good base for future work on finer-grained control of encoding. And it has demonstrated [test passes with all Python versions](https://github.com/jmarshall/pysam/actions/runs/820214292).

There's only so many balls I want to have in the air at once, and this is now holding up other improvements. @kevinjacobs-progenity: I'll aim to merge this at the end of the week if there are no objections (and if nobody else has already merged it by then).

kevinjacobs-progenity:
I've done some more light testing, reviewed the code, and am fine with merging.

PR #1007, opened 2021-04-22
Remove duplicate cdefs of PileupRead/-Column

marcelm:
The definitions in `pysam/libcalignedsegment.pxd`/`.pyx` seem to be the ones that are currently used, while these apparently are leftovers from some refactoring.


jmarshall:
Thanks. Yes, looks like they were left behind after 4edcc0a8f15528d81397080ef2e950a04c592426.

PR #1008, opened 2021-04-22
Start adding type hints

marcelm:
This is an attempt to add a couple of type hints to pysam. I’ve started with `libcalignedsegment.pyi`. I’m not finished, but need to stop for the moment.

I think these annotations would already be useful in the current state, but for this to be merged, I think we would at least need to adjust the test scripts so that they run `mypy` (it fails at the moment).


jmarshall:
Having type hints where appropriate would be good.

Is there a particular reason why you have used separate stub files rather than inline type annotations (which I believe can be used in Cython)?

Having these hints in separate files would increase the maintenance burden, as they would easily get out of sync with the definitions in the main _.pyx_ files. So IMHO for something being merged into the main pysam source project, it would be preferable to use type annotations in the main source files instead.

That would of course require dropping Python 2.7 support. That is IMHO long overdue, but it is a decision for the primary pysam maintainers (cc @AndreasHeger @kevinjacobs-progenity).

kevinjacobs-progenity:
I'm fine with dropping Python 2 support for future releases.  There are many advantages to doing so and many other commonly used bioinformatics libraries have done this.  All of my codebases have migrated to Python 3.7+ and use optional typing extensively, so I have plenty of test cases.

marcelm:
> Is there a particular reason why you have used separate stub files rather than inline type annotations (which I believe can be used in Cython)?

Using inline type annotations would be awesome, but if I’m not mistaken, it won’t actually work. Mypy can only parse annotations from a `.py` or `.pyi` file (see https://github.com/python/mypy/issues/8575#issuecomment-602188414). See also https://github.com/python/mypy/pull/8631. (But `stubgen`, on the other hand, does work somewhat – I ran it on pysam to get started with the annotations).

Also, if you define `def f(int x): pass` in a `.pyx` file, Cython generates code that checks at runtime whether the passed `x` object is an `int`. If you write this as `def f(x: int): pass`, then this code is not generated. So the runtime behavior would change. I think PEP 484 annotations are most useful (or perhaps only meant to be used) in pure Python mode.

Another problem is that I wouldn’t know how to annotate some items, for example properties.

Separate `.pyi` files have an upside as well: They allow you to present the API that you want people to use. I omitted all the deprecated functions and parameters in this PR.

Perhaps keeping the `.pyx` and `.pyi` files in sync can be solved with some additional tests.

pysam/libcbgzf.pyi
```diff
@@ -0,0 +1,40 @@
...
+from typing import Optional, Union, Any, NoReturn
+
+if sys.version_info < (3, 8):
+    from typing_extensions import Literal
+else:
+    from typing import Literal
+
+BUFFER_SIZE: int
+
+class BGZFile:
```
mr-c, line 12: Does this class implement any of the IO interfaces?

https://github.com/python/typeshed/blob/master/stdlib/io.pyi

https://github.com/python/typeshed/blob/14d7431ff1f554ae1d103b996995bfe3ecec67bd/stdlib/typing.pyi#L486

marcelm: Probably. What do you suggest?

marcelm: Type annotations for the `gzip` module are here: https://github.com/python/typeshed/blob/master/stdlib/gzip.pyi
We could copy this in the future, but I’m currently focusing on annotating the other parts of pysam.

pysam/libcbcf.pyi
```diff
@@ -0,0 +1,457 @@
...
+if sys.version_info < (3, 8):
+    from typing_extensions import Literal
+else:
+    from typing import Literal
+
+from pysam.libchtslib import HTSFile
+
+_D = TypeVar("_D")
+
+class VariantHeaderRecord:
```
kevinjacobs-progenity, line 26: Should also be Mapping[str, Optional[str]]?

pysam/libcbcf.pyi
```diff
@@ -0,0 +1,457 @@
...
+    def itervalues(self) -> Iterator[Optional[str]]: ...
+    def iteritems(self) -> Iterator[Tuple[str, Optional[str]]]: ...
+    def keys(self) -> List[str]: ...
+    def items(self) -> List[Tuple[str, Optional[str]]]: ...
+    def values(self) -> List[Optional[str]]: ...
+    def update(self, items: Union[Iterable, Dict] = ..., **kwargs) -> None: ...
+    def pop(self, key: str, default: str = ...) -> str: ...
+    # def remove(self) -> Any: ...  # crashes
+
+class VariantHeaderRecords:
```
kevinjacobs-progenity, line 53: Sequence[VariantHeaderRecord]?  May need to add ABC base class to get mixins (__contains__, __reversed__, index, count).

pysam/libcbcf.pyi
```diff
@@ -0,0 +1,457 @@
...
+    def number(self) -> Optional[str]: ...
+    @property
+    def type(self) -> Optional[str]: ...
+    @property
+    def description(self) -> Optional[str]: ...
+    @property
+    def record(self) -> Optional[VariantHeaderRecord]: ...
+    def remove_header(self) -> None: ...
+
+class VariantHeaderMetadata:
```
kevinjacobs-progenity, line 78: As above-- Mapping[str, str]?

pysam/libcbcf.pyi
```diff
@@ -0,0 +1,457 @@
...
+    def name(self) -> str: ...
+    @property
+    def id(self) -> int: ...
+    @property
+    def length(self) -> Optional[int]: ...
+    @property
+    def header_record(self) -> VariantHeaderRecord: ...
+    def remove_header(self) -> None: ...
+
+class VariantHeaderContigs:
```
kevinjacobs-progenity, line 117: As above-- Mapping[str, VariantContig]?  Should the value be Optional?

pysam/libcbcf.pyi
```diff
@@ -0,0 +1,457 @@
...
+    def iterkeys(self) -> Iterator[str]: ...
+    def itervalues(self) -> Iterator[Optional[VariantContig]]: ...
+    def iteritems(self) -> Iterator[Tuple[str, Optional[VariantContig]]]: ...
+    def keys(self) -> List[str]: ...
+    def items(self) -> List[Tuple[str, Optional[VariantContig]]]: ...
+    def values(self) -> List[Optional[VariantContig]]: ...
+    def add(self, id: str, length: Optional[int] = ..., **kwargs) -> None: ...
+
+class VariantHeaderSamples:
+    @property
```
kevinjacobs-progenity, line 137: As above -- Sequence[str]?

pysam/libcbcf.pyi
```diff
@@ -0,0 +1,457 @@
...
+    def add_line(self, line: str) -> None: ...
+    @overload
+    def add_meta(
+        self, key: str, value: None = ..., items: Iterable[Tuple[str, str]] = ...
+    ) -> None: ...
+    @overload
+    def add_meta(self, key: str, value: str = ..., items: None = ...) -> None: ...
+    def add_sample(self, name: str) -> None: ...
+
+class VariantRecordFilter:
```
kevinjacobs-progenity, line 190: As above -- Mapping[str, VariantMetadata]?

pysam/libcbcf.pyi
```diff
@@ -0,0 +1,457 @@
...
+    def iterkeys(self) -> Iterator[str]: ...
+    def itervalues(self) -> Iterator[VariantMetadata]: ...
+    def iteritems(self) -> Iterator[Tuple[str, VariantMetadata]]: ...
+    def keys(self) -> List[str]: ...
+    def items(self) -> List[Tuple[str, VariantMetadata]]: ...
+    def values(self) -> List[VariantMetadata]: ...
+    def __eq__(self, other) -> bool: ...
+    def __ne__(self, other) -> bool: ...
+
+class VariantRecordFormat:
```
kevinjacobs-progenity, line 211: As above -- Mapping[str, VariantMetadata]?

pysam/libcbcf.pyi
```diff
@@ -0,0 +1,457 @@
...
+    def iterkeys(self) -> Iterator[str]: ...
+    def itervalues(self) -> Iterator[VariantMetadata]: ...
+    def iteritems(self) -> Iterator[Tuple[str, VariantMetadata]]: ...
+    def keys(self) -> List[str]: ...
+    def items(self) -> List[Tuple[str, VariantMetadata]]: ...
+    def values(self) -> List[VariantMetadata]: ...
+
+_InfoValue = Any  # TODO see bcf_info_get_value
+
+class VariantRecordInfo:
```
kevinjacobs-progenity, line 229: Mapping[str, _InfoValue]?

pysam/libcbcf.pyi
```diff
@@ -0,0 +1,457 @@
...
+    def items(self) -> List[Tuple[str, _InfoValue]]: ...
+    def values(self) -> List[_InfoValue]: ...
+    def update(
+        self, items: Optional[Mapping[str, _InfoValue]] = ..., **kwargs
+    ) -> None: ...
+    def pop(self, key: str, default: _D = ...) -> Union[_D, _InfoValue]: ...
+    def __eq__(self, other) -> bool: ...
+    def __ne__(self, other) -> bool: ...
+
+class VariantRecordSamples:
```
kevinjacobs-progenity, line 252: Can be indexed by index and sample name, so Mapping[int | str, VariantRecordSample]?

kevinjacobs-progenity (commented):
Thanks!  This is a big job.  I've left some comments.

pysam/libcbcf.pyi
kevinjacobs-progenity: It isn't immediately clear why value should be Optional, but it has been a long time since I've dug into this code.

marcelm:
I’ve finished what I think I can do before going on vacation soon. 

This PR now covers about 80% of the API. I initially wanted to gradually add type hints, as the Mypy authors recommend, but that doesn’t work so well here because all the Cython source code is invisible to Mypy. So whenever you use anything for which no `.pyi` file exists, you get "doesn’t have attribute/module xyz" errors.

Note that the first commit in this PR is the same as PR #1009 (adding a GitHub Actions workflow). My suggestion would be to merge #1009 or some version of it separately because it’s not strictly what this PR is about.

Testing the type hints is currently done by running `mypy pysam/`. This only checks internal consistency, and should be extended at some point to also cover the `tests/` directory.

pysam/libcbcf.pyi
marcelm: Yes, this should definitely be simplified. I tried to factor out a couple of base classes, but then noticed subtle differences where the types didn’t quite match. For example, your `keys()` method returns a Python-2-style list, but `Mapping.keys()` returns an iterator. Perhaps it works anyway, but I decided to spell everything out explicitly for now. That was necessary as a first step anyway. 

The value can be None, see `VariantHeaderRecord.itervalues()`:
https://github.com/pysam-developers/pysam/blob/e340c8574694ab16fbe290c6fb714c1f6c2f5664/pysam/libcbcf.pyx#L1335-L1343

kevinjacobs-progenity: This was all written for Python 2, so it doesn't surprise me that it was never updated to the new ABCs.  Sounds like work that should be done in a follow-on PR.  Same for figuring out why keys can be None in `VariantHeaderRecord`s.

pysam/libcbcf.pyi
marcelm: Agreed, also to remaining comments. I’ll need to have a look at collections.abc to see what can be used.

pysam/libcbcf.pyi
marcelm: I have pushed an updated version that uses a custom `_Mapping` class for all the common methods.

When I try to subclass one of the ABCs like this:
```
class VariantHeaderRecord(Sized):
```
I get the error message that I haven’t implemented all abstract methods (`__len__` in this case), so it doesn’t actually help me reduce the number of lines.

marcelm:
See also the [test results](https://github.com/marcelm/pysam/actions)

pysam/libcbcf.pyi
marcelm: Do you remember why you thought the value may need to be Optional? (`__getitem__` always returns a VariantContig and never None.)

pysam/libcbcf.pyi
marcelm: I didn’t want to touch the implementation, so I haven’t actually changed this to derive from `Sequence` (the mixins are missing as you noted). That can be done at a later stage.

pysam/libcbcf.pyi
marcelm: Same here, I left this as is because VariantHeaderSamples doesn’t implement all the methods that a Sequence requires.

marcelm:
I’ve now annotated the remaining files. I added an empty `libcvcf.pyi` as it appears that `libcvcf.pyx` is outdated. Or is it?

The only remaining problem that I get now when running `mypy pysam/` is this:
```
pysam/__init__.py:26: error: Incompatible import of "TabixIterator" (imported name has type "Type[pysam.libcbcf.TabixIterator]", local name has type "Type[pysam.libctabix.TabixIterator]")
```
And this is potentially the first "bug" that Mypy has found: There’s indeed a `TabixIterator` both in `libcbcf.pyx` and `libctabix.pyx`. I’m not sure what to do, they cannot both be exposed beneath the `pysam.` namespace.

marcelm:
I took the liberty of publishing a [pysam-stubs package to PyPI](https://pypi.org/project/pysam-stubs/) for those who want to easily test the type annotations.

AndreasHeger:
Thanks! I will work on removing python2 next.

PR #1010, opened 2021-04-25
Fix typo

marcelm:
`closedf` should be `closefd`

I’m mainly working on the type annotations. Is it ok to submit tiny PRs like this for anything unrelated that I find along the way? Or would you like bigger PRs?

### DIFF
--- a/pysam/libchtslib.pyx
+++ b/pysam/libchtslib.pyx
@@ -72,7 +72,7 @@ cdef class HFile(object):
     cdef hFILE *fp
     cdef readonly object name, mode
 
-    def __init__(self, name, mode='r', closedf=True):
+    def __init__(self, name, mode='r', closefd=True):
         self._open(name, mode, closefd=True)
 
     def __dealloc__(self):


kevinjacobs-progenity:
Small is fine.  Thanks!

jmarshall:
[Personally I would prefer minor things like this to be batched up a bit. That reduces the overall effort of reviewing and merging, and there's little point in making two commits for every typo fix.]

However there's at least three problems with this code:

1. The typo in `def __init__(…, closedf=True):`, as now corrected
2. However `__init__` ignores any value provided anyway, as it always passes `True` to `_open`
3. But `_open` ignores its `closefd` parameter anyway

There are no tests, so it's difficult to guess what `closefd` was intended to mean. Perhaps something about automatically closing the incoming fd in the case that `name` is an integer? (Or mostly equivalently, not dupping it?)

kevinjacobs-progenity:
Sorry-- I was too quick on the trigger for merging.

marcelm:
Sorry, my attention appears to have been only on function headers at the time.

The `__init__` function hasn’t changed since the `HFile` wrapper was added in 4321aebd338f313b861fb1c075c64749e1e0be21. @kevinjacobs-progenity perhaps you remember why you added that parameter?

Also, (perhaps for the same underlying reason) I believe this `dup` call is never executed because `self.fp` should always be `NULL`:

https://github.com/pysam-developers/pysam/blob/15a627720fe63b9ffc310cf2f91dfc83ef795cfe/pysam/libchtslib.pyx#L92-L93

---
> Personally I would prefer minor things like this to be batched up a bit. That reduces the overall effort of reviewing and merging, and there's little point in making two commits for every typo fix.

Since you two answered differently, I now feel free to send in PRs of any size I like! :smile:

PR #1012, opened 2021-04-26
Ensure "python setup.py sdist" cythonizes extension modules

marcelm:
I saw that it is currently necessary to run `python setup.py build` before `sdist` to ensure that the generated tarball contains cythonized sources. This means that generating the tarball creates a couple of minutes.

With this PR, `python setup.py sdist` will cythonize the sources before bundling the tarball. This takes only a couple of seconds.

My intention is to use this in the CI tests: It would be nice to generate the tarball in a separate job (with the development dependencies installed), and then start a new job to install it (without developer dependencies installed). This more closely resembles what happens on a user’s system.

If Cython isn’t installed, `python setup.py sdist` will raise an ImportError when running `python setup.py sdist`. IMO the resulting error message (`ModuleNotFoundError: No module named 'Cython'`) is clear enough, but I could add a `if HAVE_CYTHON` guard if you think this would be better.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,6 +31,8 @@ import sysconfig
 from contextlib import contextmanager
 from distutils import log
 from setuptools import setup, Command
+from setuptools.command.sdist import sdist
+
 from cy_build import CyExtension as Extension, cy_build_ext as build_ext
 try:
     import cython
@@ -196,6 +198,14 @@ def get_pysam_version():
     return version.__version__
 
 
+# Override sdist command to ensure Cythonized *.c files are included.
+class cythonize_sdist(sdist):
+    def run(self):
+        from Cython.Build import cythonize
+        cythonize(self.distribution.ext_modules)
+        super().run()
+
+
 class clean_ext(Command):
     description = "clean up Cython temporary files"
     user_options = []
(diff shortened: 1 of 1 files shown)


jmarshall:
Something like this is probably a good idea if it is desired that `sdist` produces a tarball containing Cython _pysam/*.c_ files.

However I'm not convinced this is the appropriate approach for the basic CI. IMHO the natural CI to run on pushes to **master** and PRs is a matrix of `pythonX.Y setup.py build && pytest` jobs.

Producing a tarball and ensuring that that builds well is a different category of CI; one that would naturally run primarily when tags are created.

marcelm:
Sure, let’s discuss what to do in which CI job separately. Even if this only is used in a "deploy" job (or just on the command line for that matter), it improves the situation.

jmarshall:
It took me a while to understand what was going on here and why it's good to unconditionally rerun the cythonisation during `sdist`, but the penny finally dropped. By then I had simplified it to not need the global `extensions` variable. Thanks!

marcelm:
Thanks for merging, and avoiding the global is much better :-).

PR #1025, opened 2021-06-28
Remove pysam.py as it doesn’t fulfill its purpose

marcelm:
When running `python3 -c "import pysam"` in the repository root, "pysam.py" is actually ignored because `pysam/__init__.py` exists and is used instead.

I think the risk of a regular user trying to use pysam from a Git clone are quite low nowadays. Also, running `python3 setup.py build_ext -i` actually allows `import pysam` from the repo root to work.


jmarshall:
Yes, it doesn't appear to do the trick with any even vaguely pysam-supported version of Python.

Issue #1029, opened 2021-07-07
Add AArch64 wheel support


odidev:
Problem summary-

On Linux AArch64 machine, installing pysam via pip tries to build and install from the source code. It takes significant time to install it on the AArch64 user machine also user needs to have the development environment ready to install and use it on his system. Overall, we can speedup the installation time and avoid development environment dependency by release Linux.

Resolution-
`pip install pysam` should be able to download AArch64 wheels from PyPi and install it.

@jmarshall, please let me know your interest, I can help in this.

jmarshall:
Travis has mostly gone away, so we will have to redo wheel production in a different way when the time comes to make the next release — probably using GitHub Actions. So I have closed the PR, but we will certainly want to produce AArch64 builds too and your example of that will be useful when the time comes.

AndreasHeger:
I have been trying to build aarch64 packages, but have not yet found a way to do so via github actions.
Will release. If you have a wheel, happy to upload it to pypi manually.

arogozhnikov:
Not an expert, but could this help you: https://github.com/pypa/cibuildwheel ? Tools claims to be able to build wheels from github actions by using emulation

odidev:
We need to setup the qemu emulation as additional action. Please add below lines to your work flow before building the wheel-

    - name: Set up QEMU
      uses: docker/setup-qemu-action@v1

If you are interested, I can raise a PR.

PR #1030, opened 2021-07-07
Add aarch64 wheel support

odidev:
Closes https://github.com/pysam-developers/pysam/issues/1029

@jmarshall, can you please take a look into this PR?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,16 @@ _cibw_linux: &cibw_linux
     - docker
   <<: *cibw_common
 
+_cibw_linux_aarch64: &cibw_linux_aarch64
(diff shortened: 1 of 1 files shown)


odidev:
@jmarshall, can you please take a look on this PR?

jmarshall:
Travis has mostly gone away, so this PR is not directly applicable. But we will want AArch64 support, so I'll leave the issue open to track that.

jmarshall:
I was planning to delete _.travis.yml_ soon, but instead let's keep it around for reference purposes for a while even though it won't actually be activated. So I'll just merge this after all. Thanks1

Issue #1019, opened 2021-06-02
Documentation fails to build with sphinx 4


dschwoerer:
```
+ sphinx-build-3 doc html
Running Sphinx v4.0.2
Configuration error:
There is a syntax error in your configuration file: invalid syntax (conf.py, line 204)
error: Bad exit status from /var/tmp/rpm-tmp.yTQ0sN (%build)
```

PR #1031, opened 2021-07-11
Fix build with sphinx 4.0.2

dschwoerer:
Without, I get this error:
```
+ sphinx-build-3 doc html
Running Sphinx v4.0.2
Configuration error:
There is a syntax error in your configuration file: invalid syntax (conf.py, line 204)
```

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -201,8 +201,8 @@ htmlhelp_basename = 'samtoolsdoc'
 # Grouping the document tree into LaTeX files. List of tuples
 # (source start file, target name, title, author, documentclass [howto/manual]).
 latex_documents = [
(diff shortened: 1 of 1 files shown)


jmarshall:
Current **master** works for me with sphinx 4.0.2, and even after your PR _doc/conf.py_ still contains `u'foo'` strings elsewhere.

Are you sure your checkout is up to date with pysam-developers/pysam **master**? I suspect the file you are actually using contained `ur'pysam documentation'` because it did not contain e340c8574694ab16fbe290c6fb714c1f6c2f5664 which fixed #1019.

dschwoerer:
Yes, my bad. I have added e340c85 and pushed, but hadn't rebuild -.-

Would a new release make sense? It's bin over a year since the last ...

PR #1035, opened 2021-07-27
Better handling of file iteration errors

abjonnes:
For the iteration methods backed internally by `sam_read1`, _all_ return
codes < -1 signify errors which should raise an exception to the user.

***

I recently had a BAM file that had been corrupted somehow during download, but when I iterated over it with `pysam.Samfile`, iteration ended prematurely but silently - that is, no exceptions were thrown when the corrupted region of the BAM was reached, only some warnings were printed to the terminal by htslib. Example output:
```
>>> f = pysam.Samfile("mapped.bam", "rb") ; sum(1 for _ in f)
[E::idx_find_and_load] Could not retrieve index file for 'mapped.bam'
37954048
>>> f = pysam.Samfile("mapped_corrupt.bam", "rb") ; sum(1 for _ in f)
[E::idx_find_and_load] Could not retrieve index file for 'mapped_corrupt.bam'
[E::bgzf_uncompress] CRC32 checksum mismatch
[E::bgzf_read] Read block operation failed with error 33 after 112 of 149 bytes
335960
>>> 
```

This PR adds better error handling in the case of BAM files that are corrupted in ways other than truncation. The underlying [`sam_read1` method](https://github.com/samtools/htslib/blob/f7975f1ac0f55285d53cb4e72424f44f01e70981/htslib/sam.h#L1394-L1399) uses a number of different return codes less than `-1` to indicate error conditions, so the changes here raise `IOError`s if such return codes are encountered. Updated output:
```
>>> f = pysam.Samfile("mapped.bam", "rb") ; sum(1 for _ in f)
[E::idx_find_and_load] Could not retrieve index file for 'mapped.bam'
37954048
>>> f = pysam.Samfile("mapped_corrupt.bam", "rb") ; sum(1 for _ in f)
[E::idx_find_and_load] Could not retrieve index file for 'mapped_corrupt.bam'
[E::bgzf_uncompress] CRC32 checksum mismatch
[E::bgzf_read] Read block operation failed with error 33 after 112 of 149 bytes
  File "<stdin>", line 1, in <genexpr>
  File "pysam/libcalignmentfile.pyx", line 1863, in pysam.libcalignmentfile.AlignmentFile.__next__
OSError: error while reading file: -4
>>> 
```

I also included a script to generate, through the pysam test data Makefile, a BAM file which is corrupted in such a way to reproduce the issue. I can simply check in the BAM file itself if you prefer to not add the clutter of a script.

Thanks for all your great work on this library!

### DIFF
--- a/pysam/libcalignmentfile.pyx
+++ b/pysam/libcalignmentfile.pyx
@@ -1853,12 +1853,14 @@ cdef class AlignmentFile(HTSFile):
 
     def __next__(self):
         cdef int ret = self.cnext()
-        if (ret >= 0):
+        if ret >= 0:
             return makeAlignedSegment(self.b, self.header)
+        elif ret == -1:
+            raise StopIteration
         elif ret == -2:
             raise IOError('truncated file')
         else:
-            raise StopIteration
+            raise IOError("error while reading file: {}".format(ret))
 
     ###########################################
     # methods/properties referencing the header
@@ -2145,10 +2147,12 @@ cdef class IteratorRowHead(IteratorRow):
         if ret >= 0:
             self.current_row += 1
             return makeAlignedSegment(self.b, self.header)
+        elif ret == -1:
+            raise StopIteration
         elif ret == -2:
             raise IOError('truncated file')
         else:
-            raise StopIteration
+            raise IOError("error while reading file: {}".format(ret))
 
 
 cdef class IteratorRowAll(IteratorRow):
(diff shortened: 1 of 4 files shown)


rjlasko:
I hope this gets approved soon, as it will improve the reliability of `pysam` and integrity of the result.

jmarshall:
Thanks for pointing out this bug and proposing a fix. You are quite right that this code should only be raising `StopIteration` for a return code of exactly -1. The basic changes to the Python code are good, though I plan to refactor it a little.

I'd prefer to include the recipe to generate the BAM file as you have done, rather than just check the resulting binary in. It'd be nice if we could build the script out of some more general-purpose components, which I am musing about at the moment.

jmarshall:
Thanks for the improvements.

I simplified the test data file to just append a broken alignment record, which causes a -4 return code similarly to the output from _make_corrupt_bam.sh_. We may wish to resurrect _make_corrupt_bam.sh_ or an equivalent at some point to test that more complex scenario too, on this or other code paths.

PR #1039, opened 2021-09-10
Make "API" a top-level heading

marcelm:
For me, the pysam API is the most important part of the documentation, and I think the "API" section should therefore be at the top level.

I’m suggesting this change because I never seem to be able to find that section on first attempt when browsing the docs, mostly because it is somewhat hidden below the first heading "pysam - An interface for reading and writing SAM files".

While writing this, I realized that there is actually a full table of contents on the landing page https://pysam.readthedocs.io/en/latest/, which would allow me to go to the API section with a single click, but the fact that I haven’t seen it at least a dozen times makes me believe that having it also in menu on the left would be an improvement.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -2,9 +2,6 @@
 pysam - An interface for reading and writing SAM files
 ======================================================
 
-Introduction
-============
-
 Pysam is a python module that makes it easy to read and manipulate
 mapped short read sequence data stored in SAM/BAM files.  It is a
 lightweight wrapper of the htslib_ C-API.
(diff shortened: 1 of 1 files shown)


marcelm:
The second commit is unrelated, I can revert if desired.

jmarshall:
I totally agree — I don't see it to such an extent that I invariably assume it's not there and go and read the raw docstrings embedded in the _.pyx_ files instead.

I think it's further improved by having the first section stay as “Introduction” rather than “pysam - An interface for reading and writing SAM files”, so I've merged this with that change.

Issue #1044, opened 2021-09-17
Pip install fails with missing "os/rand.c"


erikbgithub:
I'm trying to pip install the package on Fedora 34 with Python 3 in a virtualenv.

While building `gcc [...] -o build/temp.linux-x86_64-3.9/htslib/hts_os.o [...]` I run into the following error:
```
    htslib/hts_os.c:30:10: fatal error: os/rand.c: No such file or directory
       30 | #include "os/rand.c"
          |          ^~~~~~~~~~~
    compilation terminated.
    error: command '/usr/bin/gcc' failed with exit code 1
```

I tried to install several packages that I use in my debian stack, where installation worked fine. But neither seems to satisfy this requirement.

```
sudo yum install htslib
sudo yum install htslib-tools
sudo yum install gcc
sudo yum install zlib-devel
sudo yum install ncurses-devel ncurses
sudo yum install Cython
sudo yum install htslib-devel
sudo yum install zlib zlib-static
sudo yum install xz-devel
sudo yum install cryptopp
sudo yum install ncurses-devel ncurses
sudo yum install gsl-devel gsl
sudo yum install perl-devel perl
sudo yum install BEDTools                          <- sadly doesn't exist anymore
sudo yum install libstdc++-devel libstdc++
sudo yum install glibc
sudo yum install bzlib bzlib-devel
sudo yum install gambas3-gb-compress-bzlib2
sudo yum install lbzip2 bzip2 lbzip2-utils
```

erikbgithub:
On my debian 11 systems I install this:

```
sudo apt update
sudo apt upgrade -y
sudo apt install \
  git vim make tmux zip bc pigz rsync\
  apache2-utils \
  sshfs \
  clang gcc autoconf \
  zlib1g zlib1g-dev \
  libbz2-dev libbz2-1.0 \
  liblzma5 liblzma-dev \
  libcurl4 libcurl4-gnutls-dev \
  libcrypto++8 libcrypto++-dev \
  libncurses5-dev libncursesw5-dev \
  libgsl25 libgsl-dev \
  libperl-dev \
  bedtools \
  python3-dev \
  pyvcf \
  python3 \
  default-jre \
  python3-venv \
  python3-pysam \
  python3-pyvcf \
  python3-numpy
```

And there I also manually compile all the libs (htslib, samtools, bcftools, bwa, minimap2, bwa-mem2)

jmarshall:
[...]

PR #1045, opened 2021-09-17
baseimage for containers

erikbgithub:
After running into #1044 I've started exploring this path. And here the installation "just worked". If we provide a base image with the pysam package then people can build their tooling on top of that baseimage in any OS that runs containers in a similar fashion.

I built and tested it this way:

```
$ podman build -t pysam -f Containerfile .
$ podman run -it pysam python -c "import pysam; print('woohoo')"
```

Let me know what you think.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,0 +1,13 @@
+FROM ubi8:latest
+
+RUN yum update \
+  && yum install -y python3-pip python3-devel pigz \
+  && cd /usr/local/bin \
+  && ln -s /usr/bin/python3 python \
+  && pip3 --no-cache-dir install --upgrade pip \
+  && yum clean all \
+  && echo "system packages installed"
+
+RUN python -m pip install pysam
+
+WORKDIR /opt/


AndreasHeger:
Hi @erikbgithub 

Thanks. Happy to add it if it makes it easiers for users to install.

jmarshall:
Apologies for not commenting out loud on this one previously.

The OP's problem was sorted out in #1044, and as @marcelm commented there, the problem is really solved by providing wheels.

The _Containerfile_ suggested appears to be specific to a particular Red Hat ecosystem (“ubi8”) and includes some infelicities: why does it install `pigz`? symlinking `python` is probably not a great idea. I'm not sure it'll be particularly helpful to anyone, including the OP.

erikbgithub:
@AndreasHeger @jmarshall Hi. No problem for the late response.

Usually, I don't post a completely finished PRs, but already an early draft version expecting feedback, questions and strategic alignments. So when I get to finish things up, then I can already include the feedback from other people that I wouldn't have expected otherwise.

In other words: Your criticism is welcome and next time discuss it before clicking the merge button. Otherwise, I'm not satisfied either.

If you consider removing the Containerfile again in its current state, I wouldn't mind.

AndreasHeger:
@erikbgithub and @jmarshall , thanks. 

I will clean it out again.

PR #1061, opened 2021-11-15
Base mods

kpalin:
Partial interface to htslib modified base interface. Currently only returns query and reference coordinates of an AlignedSegment. No pileup or point query interaces.

### DIFF
--- a/pysam/libcalignedsegment.pyx
+++ b/pysam/libcalignedsegment.pyx
@@ -1747,7 +1747,7 @@ cdef class AlignedSegment:
         """Modified bases annotations from Ml/Mm tags. The output is
         Dict[(canonical base, strand, modification)] -> [ (pos,qual), ...]
         with qual being (256*probability), or -1 if unknown.
-        Strand==0 for forward and 1 for reverse strand modification
+        Strand=='+' for forward and '-' for reverse strand modification
         """
         def __get__(self):
             cdef bam1_t * src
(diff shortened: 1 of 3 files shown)


rrazaghi:
Is there a plan to merge this? or fully support modified bases in pysam?
This would be a huge help for long-read tool development!
Thanks!

jmarshall:
These API functions were new in HTSlib 1.14. AIUI pysam tries to be compatible with a range of HTSlib versions, so starting to use such recent API functions would require a policy decision.

On the other hand, as pysam imports recent versions of samtools & bcftools and these generally require correspondingly recent versions of htslib, it's possible that pysam as a whole is only compatible with recent htslib after all.

Either way, it would be good to have a ruling from the maintainers on the direction and policy here.

AndreasHeger:
Thanks. pysam generally used to be tied to a particular htslib version. It might only recently have been the case that it became compatible with multiple htslib versions, most likely due to lack of new feature-development on pysam's side (for which I apologize). In my view, providing access to new htslib features take precedence over backwards compatibility to older htslib versions.

AndreasHeger:
Many thanks, @kpalin , merging now.

AndreasHeger:
I merged your changes manually as I wanted to run tests locally, but this seems not get recorded here. You are still credited as author, though, so hope that is all ok.
Many thanks.

I noticed you referenced some test data files (MM-pileup1.sam and MM-pileup2.sam). Did you create these locally?

AndreasHeger:
Closing for now.

PR #1063, opened 2021-11-26
Run flake8 and fix some linting issues

marcelm:
Whenever I open up the pysam source code in PyCharm, I get a lot of warnings from the built-in code style checker: Unused imports, weird indentation, too many blank lines, trailing whitespace etc. ... I have somewhat gotten used to working with code that uses an automatic style checker, mostly flake8, so I wonder if this would be something that could be considered for pysam.

In this PR, I’ve
- added a "lint" job to the GitHub Action workflow that runs `flake8`
- configured `flake8` to ignore all the existing problems it finds
- re-enabled a couple of these checks while at the same time fixing them in the source code

The idea is to re-enable the checks one by one. I can do this for some of the checks if there’s interest.

I’m aware that style checking is a matter of personal preference, and perhaps it’s not desirable to have CI fail "just" because of style issues, so I’m perfectly ok if this doesn’t get accepted.

(I use Black in other projects, but didn’t want to bring this up, yet, but it’s of course another option.)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,7 +14,7 @@ jobs:
       - name: Install dependencies
         run: python -m pip install flake8
       - name: Run flake8
-        run: flake8 pysam/ tests/
+        run: flake8 pysam/ tests/ *.py
 
   direct:
     runs-on: ${{ matrix.os }}-latest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,6 +3,19 @@ name: CI
 on: [push, pull_request]
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: python -m pip install flake8
+      - name: Run flake8
+        run: flake8 pysam/ tests/
+
   direct:
     runs-on: ${{ matrix.os }}-latest
     strategy:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,19 +3,6 @@ name: CI
 on: [push, pull_request]
 
 jobs:
-  lint:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - name: Set up Python
-        uses: actions/setup-python@v2
-        with:
-          python-version: "3.10"
-      - name: Install dependencies
-        run: python -m pip install flake8
-      - name: Run flake8
-        run: flake8 pysam/ tests/ *.py
-
   direct:
     runs-on: ${{ matrix.os }}-latest
     strategy:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,10 +7,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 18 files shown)


AndreasHeger:
Thanks, @marcelm , I am very much in favour of pep8 and automatic linting as part of CI.

.github/workflows/ci.yaml
```diff
@@ -3,6 +3,19 @@ name: CI
 on: [push, pull_request]
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
```
jmarshall, line 10: This job doesn't have a `matrix`, so `matrix.python-version` is empty. You appear to have got the default for `actions/setup-python@v2` but it's not very clear what its default is.

Probably better to hardcode 3.10 or something like that.

marcelm: You’re right; I copied this from another project which had a matrix. Hardcoded to Python 3.10 now.

marcelm:
Thanks for the encouragement :-). I pushed a few more commits that fix other things flake8 complains about and also added `setup.py` and `cy_build.py` to the list of files to check.

setup.py
```diff
@@ -445,7 +445,8 @@ def run(self):
 # reasons of simplicity.
 
 def prebuild_libchtslib(ext, force):
-    if HTSLIB_MODE not in ['shared', 'separate']: return
+    if HTSLIB_MODE not in ['shared', 'separate']:
+        return
```
jmarshall, line 450: I believe judicious use of single-line conditionals improves readability, and I would thank you to leave instances like this as they are.

If E701 warns for lines like this, I suggest you leave E701 deactivated. Note that PEP-8 admits that single-line conditionals are sometimes appropriate.

setup.py
```diff
@@ -444,13 +444,17 @@ def run(self):
...
 # reasons of simplicity.
 
+
 def prebuild_libchtslib(ext, force):
-    if HTSLIB_MODE not in ['shared', 'separate']: return
+    if HTSLIB_MODE not in ['shared', 'separate']:
+        return
     write_configvars_header("htslib/config_vars.h", ext, "HTS")
 
+
```
jmarshall, line 453: For small functions, this is counterproductive.

I suggest you leave the most prescriptive number-of-blank-lines rules deactivated.

jmarshall:
As a counterpoint: Automatic linting against unfamiliar rules will make contributing to pysam harder for those who are not intimately familiar with these rules. I count myself amongst that category, and I am not in favour of things that make it harder for me to contribute to pysam. :smile:

Some of this is reasonable and suitable for automatic linting; a passing familiarity with PEP-8 and typical Python code layout is fair to assume from contributors. Some other rules are arbitrary and prescriptivist, serve mainly to discourage contributors without substantially (or at all in some cases) improving readability, and IMHO pysam would do well to leave them deactivated.

marcelm:
I’ve backed out the additional changes. I have left in my changes fixing "F401 '...' imported but unused" and "E111 indentation is not a multiple of 4".

I am also totally fine with removing the CI job if one of you is worried about current or future contributors being annoyed by CI complaints.

AndreasHeger:
Thanks both. These are all valid points. 
We can run the linting CI job as a report initially, but it is good to have the infrastructure in place now.

marcelm:
Unfortunately, it’s not possible to run a GitHub Actions job and let its result be ignored, so I have removed the "lint" job. The benefit of this PR is now a one-time fix of a couple of style issues and unused imports and addition of flake8 configuration settings.

AndreasHeger:
Ok, thanks.

PR #1078, opened 2022-02-05
Support BAM_CPAD in get_aligned_pairs

### LINKED ISSUE #1077: AlignedSegment.get_aligned_pairs raises NotImplementedError on BAM_CPAD
When `get_aligned_pairs` it's an unfortunate show-stopper for those trying to make use of that function. Looking at the code leading up to line 1977 of `libcalignedsegment.pyx`:

```python
            elif op == BAM_CPAD:
                raise NotImplementedError(
                    "Padding (BAM_CPAD, 6) is currently not supported. "
                    "Please implement. Sorry about that.")
```

it looks like a simple modification would be to just append `(None, None)` to `result` in this case (or `(None, None, None)` if `_with_seq`), document it, and let the caller deal with the situation.

As the code stands right now, the caller gets nothing back at all, just an exception.

I could probably make this change and add tests. Would this be considered a reasonable "fix" here?

This is related to #1076 which was opened eight days ago (and then closed after three days with no comments) by @DadongZ

terrycojones:
Hi. This isn't ready to be merged, I'm just making the first step towards making handling of `BAM_CPAD` not throw an exception in `get_aligned_pairs` so that people who use that function can continue to call it and deal with the result themselves.  But I don't know my way around the pysam codebase at all, so would like a little help, please! :-)

I don't know how to rebuild `pysam/libcalignedsegment.pxd` in order to run the tests on the new code. I tried `python setup.py build_ext --inplace` but it built me a `pysam/libcalignedsegment.cpython-310-darwin.so` and when I run the test the old behavior is still in place. I moved the corresponding `.pxd` file aside, but that didn't help.

I'm not sure if my tests are correct, but will be happy to fix them if not.

Thanks!  All this is on the assumption that what I suggested in #1077 seems like an ok approach.

Fixes #1077


AndreasHeger:
Thanks, @terrycojones , this looks good to me. 
I will add some explanation to the documentation.

terrycojones:
Hi @AndreasHeger - thanks for the reply. Do you mean you'll add some explanation on how to rebuild & run all tests, etc? Or add to the docs to mention this change in the behavior of the function? I'm happy to help out further in either case.

AndreasHeger:
Sorry, I was not very clear, I meant to add some text to the docstring.

The .pxd files are manually created and contain declarations, similar to header files (.h) in C. They need not be re-generated. To rebuild, I usually do a:

AndreasHeger:
I have also enabled the CI tests. 

If you find the time to look at the tests, that would be great, and if you could add an explanation to the docstrings, that would also be good.

terrycojones:
OK, I've made some progress. My reading of the SAM spec (section 3.1) is that when a CIGAR string contains a `P` we are dealing with a situation in which some tool has inserted (normally) `*` chars into the reference sequence in order to make it simple to specify insertions to the reference using `P` in the CIGAR string.  This is different from "Padded SAM" (section 3.2) in which the `P` operation is not used.

In other words, this is just an insertion into the reference, though specified in a different way.  So I have modified the pull request to mirror what is done when there is an insertion to the reference. One can tell the difference (if needed), either by looking at the CIGAR operations or by passing `with_seq=True` to `get_aligned_pairs` (in which case a `*` comes back as the reference character, following the convention suggested by the SAM spec). `None` could also be returned as the reference character in this case. I don't see any harm in providing this information via `*` but strictly speaking it's not needed since the caller can always see the `P` in the CIGAR string if they really care.

BTW, I put a call to `free(s)` into `build_alignment_sequence` right before a raise of an `AssertionError`. It's unlikely, but a caller could write a loop that repeatedly calls `build_alignment_sequence` and ignores the exception and there would be a memory leak. I can remove the `free(s)` call if you like.

terrycojones:
And thanks for the help on rebuilding!

terrycojones:
In the light of day, I now think it would be more consistent to use `None` for the reference, not `*`. Then `get_aligned_pairs` could treat `BAM_CPAD` exactly as it already treats `BAM_INS` and `BAM_CSOFT_CLIP` because this really is an insertion. Another argument for treating them in the same way is that existing (and future) code calling `get_aligned_pairs` will just work and be simpler. There may be existing code that currently catches and ignores a `NotImplementedError` that will now not occur - instead the flow of control will pass through the insertion logic, as it should do. So the change in this pull request could cause existing code to behave differently (e.g., to produce a little more output because a formerly unsupported CIGAR string is now supported). I hope I'm making sense...

terrycojones:
Any opinion on this? Shall I switch to using `None`?  This is non-urgent for me, so no rush, just wondering.

AndreasHeger:
Hi @terrycojones , many thanks for this and your great explanations. I agree that using `None` sounds much simpler and avoids having a user check for `*`.

terrycojones:
Hi @AndreasHeger OK, done. Sorry for the slowness - I redirect GitHub emails to folder I rarely look at...  I hope this works!

AndreasHeger:
Many thanks!

terrycojones:
@AndreasHeger and thanks to you!

PR #1080, opened 2022-02-13
No longer refer to deprecated attributes in AlignedSegment attribute descriptions

marcelm:
Some of the attribute docstrings refer to deprecated attributes. For example, `reference_length` is described as being equal to `aend - pos`. This PR fixes those outdated descriptions.


AndreasHeger:
Many thanks!

PR #1083, opened 2022-03-03
Update imports to htslib/samtools/bcftools 1.15 and resolve symbol conflicts

### LINKED ISSUE #1081: Potentially OS-specific `bcftools.mpileup()` error: `Segmentation fault (core dumped)`
Hello developers of `pysam`,

As always, thanks for your hard work! Today, I'm reporting a potential OS-specific bug in `bcftools.mpileup()` where the method keeps killing the running program with `Segmentation fault (core dumped)`. The simplest example to reproduce the error is:

```
$ cat pysam_test.py
from pysam import bcftools
bcftools.mpileup('in.bam', '-f', 'ref.fa')
```

```
$ python3 pysam_test.py
Segmentation fault (core dumped)
```

In the example above, I put `'in.bam'` and `'ref.fa'` instead of actual filenames because I've tried a number of different files and they all gave me the same error.

Interestingly, this error only occurs in linux server; it works just fine in macOS. That's why I think this might be a OS-specific bug. Below is my linux server:

```
$ uname -a
Linux bdcm03 3.10.0-514.2.2.el7.x86_64 #1 SMP Tue Dec 6 23:06:41 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```

And my macOS:

```
$ uname -a
Darwin Seung-beens-Mac-mini.local 20.3.0 Darwin Kernel Version 20.3.0: Thu Jan 21 00:06:51 PST 2021; root:xnu-7195.81.3~1/RELEASE_ARM64_T8101 arm64
```

I'm using the latest version for both OS systems (0.18.0 for `pysam` and 1.14 for `htslib`, downloaded via conda). I checked other issues in this repository, including my previous ones (#928, #944), but could not find any answers.

Any help would be greatly appreciated!

jmarshall:
* Instead of rewriting clashing symbols like `main_reheader()` in _devtools/import.py_, `#define` them in _import/pysam.h_ so that the C preprocessor replaces them with distinct per-tool-prefixed symbol names. This is effective in both `*.c` and `*.h` files.

* Add an extra step to `python3 setup.py build` that detects any symbol conflicts, causing build failure if there are any. This is really a build error, just as one of the Cython extensions failing to build or link would be.

* Do the actual imports of the upstream 1.15 releases.

* Add more `#define` directives to _import/pysam.h_ to resolve all remaining symbol conflicts. (Since VCF/BCF mpileup was removed in samtools 1.15, there are many fewer conflicts than there were in 1.14.) This, together with failing the build if any conflicts reappear in future, fixes #1081 once and for all.

* Add 1.15's new subcommands: `samtools consensus` & `head`, and `bcftools head`.


AndreasHeger:
Hi @jmarshall , thank you so much for this. Is the MR ok to be merged?

jmarshall:
Yes, it's passed my testing and I believe is ready to be merged. You may wish to test it too :smile:

AndreasHeger:
Will do. And again, many thanks, you have again done all the heavy lifting.

PR #1085, opened 2022-03-04
Improve getopt()/getopt_long() resetting when running samtools/bcftools commands

### LINKED ISSUE #813: pysam.cat finds invalid options that were not passed
Hello, 

I have found an issue with pysam.cat(), in which I get errors saying I have used invalid options. Yet, none of the options match what was used in the call args.

Example:
```
cat_args = ['-o', '/tmp/tmp33qgrabk.cat.bam', '/tmp/tmp9b1duxnw.masked.bam', '/tmp/tmpa30gy13u.primer.nonintersected.bam']
pysam.cat(*cat_args, catch_stdout=False)
```
returns
```
cat: invalid option -- '?'
cat: invalid option -- 'U'
cat: invalid option -- 'U'
cat: invalid option -- 'U'
...
 File "/home1/06285/ihoskins/bioinfo/analysis/seq_utils.py", line 464, in cat_bams
    pysam.cat(*cat_args, catch_stdout=False)
  File "/home1/06285/ihoskins/miniconda3/envs/CBS_variants/lib/python3.6/site-packages/pysam/utils.py", line 75, in __call__
    stderr))
pysam.utils.SamtoolsError: 'samtools returned with error 1: stdout=None, stderr=[main_cat] ERROR: input is not BAM or CRAM\n'
```

```
Python 3.6.6 | packaged by conda-forge | (default, Oct 12 2018, 14:43:46) 
[GCC 7.3.0] on linux
>>> import pysam
>>> pysam.__version__
'0.15.2'
>>> pysam.__samtools_version__
'1.9'
```

I checked to verify the input files are BAM format. One of them happens to have no alignments, but I don't believe that is the issue. The problem appears to be parsing the call args, whereas the downstream error is probably caused by failure to identify the BAM files.

What is perplexing to me is that if I run the same call on the same files in an interactive python session (or if I just call from the command line with my own samtools distribution), the cat command runs just fine and produces the expected output. Furthermore, if I simply put a pdb breakpoint before the call in my script, then continue (c), the call completes without error. It also works if I simply step through the function. So, it seems the issue has to do with my environment, specifically when called from a script. I tried importing pysam just prior to the call, but that did not help.

Has anyone observed this behavior before? I recall experiencing the same type of error a few years ago with pysam.index() when specifically called from a script (and on a different architecture).

I am happy to provide my test files if that helps. Thanks in advance for any help!

### LINKED ISSUE #1084: Error ‘invalid option’ when using pysam.index after to pysam.depth
Hello,

Below the description of the bug. Thanks in advance for your explanation/solution.

# Description

After using pysam.depth with the option ‘–s’, the following calls to pysam.index raise this exception: 

```
index: invalid option -- '�'
  File "test_index.py", line 26, in <module>
    pysam.index(tmp_bam)
  File "/home/user/Documents/code/project/venv/lib/python3.8/site-packages/pysam/utils.py", line 69, in __call__
    raise SamtoolsError(
pysam.utils.SamtoolsError: 'samtools returned with error 1: stdout=, stderr=Usage: samtools index [-bc] [-m INT] <in.bam> [out.index]\nOptions:\n  -b       Generate BAI-format index for BAM files [default]\n  -c       Generate CSI-format index for BAM files\n  -m INT   Set minimum interval size for CSI indices to 2^INT [14]\n  -@ INT   Sets the number of threads [none]\n'
```

Note: The invalid character may change when the code is rerun.
No error is raised if I remove ‘-s’ option in pysam.depth.

# Pysam release

0.18.0

# Code to reproduce

```
import pysam
import os

tmp_bam = "/tmp/tmp.bam"
tmp_sam = "/tmp/tmp.sam"

# Write SAM
with open(tmp_sam, "w") as writer:
    writer.write("""@SQ	SN:chr1	LN:60
read1	0	chr1	1	60	34M	*	0	0	ATGACTGAATATAAACTTGTGGTAGTTGGAGCTG	*	NM:i:0	MD:Z:34	AS:i:34	XS:i:0
""")
# SAM to BAM
with pysam.AlignmentFile(tmp_sam, "r") as reader:
    with pysam.AlignmentFile(tmp_bam, "wb", header=reader.header) as writer:
        for record in reader:
            writer.write(record)
# Test
pysam.index(tmp_bam)
pysam.depth(
    "-a",
    "-r", "chr1:5-15",
    "-s",
    tmp_bam
)
os.remove(tmp_bam + ".bai")
pysam.index(tmp_bam)
```

jmarshall:
The `getopt()` and `getopt_long()` functions have internal state tracking the option parsing that needs to be reset when user code wants to start processing a different `argc`/`argv` pair, i.e., when running a samtools or bcftools command (apart from the first time such a command is run in a Python session).

7954ff1a6df7a4a7e088e7b17bfd4e9868bb94dc improved resetting on macOS but broke it on Linux. This PR rewrites it to work on both platforms. Resetting is not standardised by POSIX but varies by C library:

* GNU libc defines its own way by setting `optind` to 0. (This is barely documented, but is described in the _getopt.h_ header.)
* BSD-derived systems use an extra `optreset` variable, which can be detected via a preprocessor symbol: some BSDs define `_OPTRESET_DECLARED`, while macOS has defined `_OPTRESET` since 10.4 (Tiger).
* Otherwise the best attempt just sets `optind` to 1.

This PR adds those three variants to `@pysam@_dispatch()` and reruns `devtools/import.py` to regenerate the samtools/bcftools files derived from _pysam/import.c_. Fixes #813 and fixes #1084.

Because this affects the same imported files as PR #1083, the best way to apply it will be to merge #1083 first and then rebase (and re-import) this on top of the 1.15 import. Hence this PR is left as draft until then.

### DIFF
--- a/bcftools/bcftools.pysam.c
+++ b/bcftools/bcftools.pysam.c
@@ -1,5 +1,4 @@
-#include <ctype.h>
-#include <assert.h>
+#include <getopt.h>
 #include <unistd.h>
 #include <setjmp.h>
 #include <stdio.h>
@@ -62,6 +61,15 @@ static int bcftools_status = 0;
 
 int bcftools_dispatch(int argc, char *argv[])
 {
+  /* Reset getopt()/getopt_long() processing. */
+#if defined __GLIBC__
+  optind = 0;
+#elif defined _OPTRESET || defined _OPTRESET_DECLARED
+  optreset = optind = 1;
+#else
+  optind = 1;
+#endif
+
   if (setjmp(bcftools_jmpbuf) == 0)
     return bcftools_main(argc, argv);
   else
@@ -73,17 +81,3 @@ void bcftools_exit(int status)
   bcftools_status = status;
   longjmp(bcftools_jmpbuf, 1);
 }
-
-
-void bcftools_set_optind(int val)
-{
-  // setting this in cython via 
-  // "from posix.unistd cimport optind"
-  // did not work.
-  //
-  // setting to 0 forces a complete re-initialization
-  optind = val;
-}
-
-
-
--- a/bcftools/bcftools.pysam.h
+++ b/bcftools/bcftools.pysam.h
@@ -53,8 +53,6 @@ int bcftools_dispatch(int argc, char *argv[]);
 
 void PYSAM_NORETURN bcftools_exit(int status);
 
-void bcftools_set_optind(int);
-
 extern int bcftools_main(int argc, char *argv[]);
 
 /* Define these only in samtools/bcftools C source, not Cython code. */
--- a/import/pysam.c
+++ b/import/pysam.c
@@ -1,5 +1,4 @@
-#include <ctype.h>
-#include <assert.h>
+#include <getopt.h>
 #include <unistd.h>
 #include <setjmp.h>
 #include <stdio.h>
@@ -62,6 +61,15 @@ static int @pysam@_status = 0;
 
 int @pysam@_dispatch(int argc, char *argv[])
 {
+  /* Reset getopt()/getopt_long() processing. */
+#if defined __GLIBC__
+  optind = 0;
+#elif defined _OPTRESET || defined _OPTRESET_DECLARED
+  optreset = optind = 1;
+#else
+  optind = 1;
+#endif
+
   if (setjmp(@pysam@_jmpbuf) == 0)
     return @pysam@_main(argc, argv);
   else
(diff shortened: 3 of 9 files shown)


jmarshall:
I'll need to rebase and update it (after #1083 lands) before this one can be merged.

jmarshall:
@AndreasHeger: This one is updated and ready for merging now too.

AndreasHeger:
Many thanks!

PR #1088, opened 2022-03-09
Add AlignedSegment is_mapped/mate_is_mapped/is_forward/mate_is_forward properties

jmarshall:
These are opposites of the existing `is_unmapped`/`mate_is_unmapped`/`is_reverse`/`mate_is_reverse` properties. The existing properties match the sense of the underlying flag bits; these proposed additions allow more natural conditions to be written in some cases:

```python
if not aln.is_unmapped:
    print('stilted way to say if mapped')

if aln.is_mapped:
    print('more natural')

if aln.is_forward:
    print('handle the forward case first')
else:
    print('then the reverse case')
```

Inspired by a [minor grizzle by Nick Loman](https://twitter.com/pathogenomenick/status/1104006763559505920).


AndreasHeger:
Thanks, it is nice to be nice :=)

jmarshall:
Thanks!

PR #1092, opened 2022-03-30
Add VariantRecord.alleles_variant_types

ajcr:
This adds a read-only property `alleles_variant_types` to `VaraintRecord`.

In the past, I've needed a way to check variant types when processing VCF files (splitting/counting records or samples based on type). The HTSlib function `bcf_get_variant_type` can be used for this but is not currently exposed via pysam.

This new property returns a tuple containing the type of each variant in `VaraintRecord.alleles`. 

(I couldn't decide on the right name, but wanted to associate it with `VariantRecord.alleles` as it returns a tuple of the same length but with the type of each allele - happy to change it to something better.)

### DIFF
--- a/pysam/libcbcf.pyx
+++ b/pysam/libcbcf.pyx
@@ -3242,6 +3242,7 @@ cdef class VariantRecord(object):
             self.ptr.rlen = rlen
         else:
             self.ptr.rlen = len(values[0])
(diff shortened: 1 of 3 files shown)


AndreasHeger:
Thanks, very useful. Will merge.

Issue #1104, opened 2022-05-20
feat: a faster way to add samples to a VCF


aryarm:
Thanks for making `pysam`! It's such a useful library!

## My use case
I'm trying to create a VCF of 400,000 samples (and just a few variants) from python. But unfortunately, the operation appears to take a very long time.

## The issue
The first culprit appears to be `VariantHeader.add_sample()`, which does not appear to be linear.
Consider the following python snippet:
```python
from pysam import VariantHeader
vh = VariantHeader()
for i in range(num_samples):
    vh.add_sample(f"sample{i}")
```
Running this snippet under various values of `num_samples` results in the following.
![image](https://user-images.githubusercontent.com/23412689/169433331-33893fae-5e32-43d7-8300-abb3418a1bb1.png)
Extrapolating for 400,000 samples: it would probably take me ~13 hrs to run this snippet.

## A hypothesis
I'm guessing [this realloc()](https://github.com/pysam-developers/pysam/blob/be6bc7b445d5ff8ecce217143ecc7e8dd9ceee13/htslib/vcf.c#L116) is [making a new copy](https://en.cppreference.com/w/c/memory/realloc) of the list of samples every time `add_sample()` is called?

## My feature request
Can there be a way to add multiple samples at once, instead of adding them one at a time? Or could there be an option to declare the number of samples when initializing the VariantHeader object, so that the memory can be preallocated?

Apologies if there already is a way of doing this that I'm just unfamiliar with!

## Potential workarounds
One idea is to use [`pgenlib`](https://github.com/chrchang/plink-ng/tree/master/2.0/Python) to create a PLINK2 PSAM (and PGEN + PVAR) file and then use the `plink` command-line utility to convert it to VCF.

Another idea is to create an empty VCF with just a header consisting of a few samples via `pysam`, append the rest of the samples to the file via python's `open('file.vcf', 'a').write()`, and then read the VariantHeader from the file back again using `pysam` or another library.

I've tested neither of these ideas so far, though.

jmarshall:
While reallocing every time is not ideal, I suspect the real cause of this slowness is that `add_sample()` does a `bcf_hdr_sync()` (itself an O(N) operation) for every sample added, leading to O(N<sup>2</sup>) behaviour:

https://github.com/pysam-developers/pysam/blob/040ce8b5ad68de1cc291eaed94e121a41e13f254/pysam/libcbcf.pyx#L2182-L2188

[...]

PR #1105, opened 2022-05-20
New VariantHeader.add_samples() method to add multiple samples at once

jmarshall:
Issue #1104 reports that adding hundreds of thousands of samples to a VCF header via the same number of calls to `VariantHeader.add_sample()` is very slow. I suspect this is because `add_sample()` does a `bcf_hdr_sync()` (itself an O(N) operation) for every sample added, leading to O(N<sup>2</sup>) behaviour.

This PR adds a new `add_samples()` method that can add an arbitrary number of samples and then call `bcf_hdr_sync()` once. It also adds an internal wrapper for `bcf_hdr_sync` that adds error checking.

```python
def add_samples(self, *args):
    """Add several new samples to this header.
    This function takes multiple arguments, each of which may
    be either a sample name or an iterable returning sample names
    (e.g., a list of sample names).
    """
```


jmarshall:
Now tested that the new `add_samples()` function can add 400,000 samples in a second, thus confirming that doing `bcf_hdr_sync()` only once solves the reported slowness problem.

This could be done compatibly by altering the arguments of `add_sample()` instead, so that it could take one or many arguments or iterables — existing code using `foo.add_sample("bar")` would still do the same thing. But I considered that there was a difference in emphasis between add-one-sample and add-lots-of-samples that was worth reflecting in the new `add_samples()` function name.

AndreasHeger:
Thanks for this and agree that a separate function makes sense.

PR #1116, opened 2022-07-06
Fix types for variable length tuples in libcbcf.py

nh13:


pysam/libcbcf.pyi
```diff
@@ -150,7 +150,7 @@ class VariantHeader:
         contig: Optional[str] = ...,
         start: int = ...,
         stop: int = ...,
-        alleles: Optional[Tuple[str]] = ...,
+        alleles: Optional[Tuple[str, ...]] = ...,
```
nh13, line 153: I believe all of these are variable length, so need an ellipsis after: https://docs.python.org/3/library/typing.html#typing.Tuple

marcelm:
I can confirm that this is incorrect in master and should be fixed the way Nils suggests.

AndreasHeger:
Thanks

PR #1117, opened 2022-07-06
Fix typing of new_record from VariantHeder

nh13:
See: https://github.com/pysam-developers/pysam/blob/5eac009c5e85d280e8bbcb69ed0fa081e2271d78/pysam/libcbcf.pyx#L2120-L2124


AndreasHeger:
Thanks

PR #1118, opened 2022-07-06
Fix type of VariantHeaderMetadata add

nh13:
Gets made into a string here:
https://github.com/pysam-developers/pysam/blob/5eac009c5e85d280e8bbcb69ed0fa081e2271d78/pysam/libcbcf.pyx#L1587


AndreasHeger:
Thanks

Issue #1129, opened 2022-09-30
AlignmentFile.pileup should return PileupColumns


mpcusack-color:
I might be missing something, but it seems like the `AlignmentFile.pileup` function is type annotated to return the wrong types. From the code and documentation they appear to be `PileupColumn`s, but the type hints say they are `AlignedSegment`s.

- `AlignmentFile.pileup` returns a subclass of `IteratorColumn`: https://github.com/pysam-developers/pysam/blob/56b1624c21b02762fdcea3aa5f657a4152ca178e/pysam/libcalignmentfile.pyx#L1351
- `IteratorColumn`s return `PileupColumn` objects: https://github.com/pysam-developers/pysam/blob/56b1624c21b02762fdcea3aa5f657a4152ca178e/pysam/libcalignmentfile.pyx#L2814
- `IteratorColumn`s are annotated to return `AlignedSegment`s: https://github.com/pysam-developers/pysam/blob/56b1624c21b02762fdcea3aa5f657a4152ca178e/pysam/libcalignmentfile.pyi#L204

jmarshall:
I think perhaps the problems are even a bit more than that. e.g.:

* The `makePileupColumn()` call on line 2814 is in fact in `class IteratorColumnAll`, but that's not annotated
* `IteratorColumn`'s `__iter__()` method surely returns an `IteratorColumn`

Would you be interested in auditing these classes' annotations further?

PR #1130, opened 2022-09-30
Fix type hints for AlignmentFile.pileup to return PileupColumns

mpcusack-color:
Fixes https://github.com/pysam-developers/pysam/issues/1129

As far as I can tell this was just a oversight when the type hints were added, and there isn't anything more complicated going on (e.g. it returns different types at different time).


pysam/libcalignmentfile.pyi
```diff
@@ -201,7 +201,7 @@ class IteratorRowSelection(IteratorRow): ...
 
 class IteratorColumn:
     def __iter__(self) -> IteratorRow: ...
```
marcelm, line 203: While you are here anyway, can you fix this to `IteratorColumn` as @jmarshall noted in https://github.com/pysam-developers/pysam/issues/1129#issuecomment-1264739672?
(I added these type annotations originally and can confirm that this was just an oversight. In this case probably a copy&paste error.)

PR #1134, opened 2022-10-13
Fix get_tag and get_tags type hints

marcelm:
Without this, calling `get_tag()` and omitting `with_value_type` is flagged as a type error.


AndreasHeger:
Thanks

Issue #1112, opened 2022-06-20
Some htscodecs objects will soon need some target-specific compiler flags


mp15:
The current version of htslib uses a version of htscodecs which uses autoconf, automake, and libtool. devtools/import.py does not copy enough of the supporting files across, resulting in a spurious error when building on certain systems:

`error: inlining failed in call to ‘always_inline’ ‘_mm256_insertf128_si256’: target specific option mismatch`

This error is caused by the default makefile fall through (and thus the march has not been set correctly for this target).

jmarshall:
There is no problem in current pysam, which uses the current HTSlib release — version 1.15.1, which does not include any files needing this treatment. However when we come to import a future release of HTSlib that does include this stuff, the pysam build will fail. Presumably this is what you are doing — trying to import current **develop** from HTSlib.

Ugh, who thought it would be a good idea to use automake or libtool? 😛 🤣

But that's a red herring for us anyway: pysam builds htscodecs using the bundled build rules in **HTSlib's _Makefile_** and does not use htscodecs's own build infrastructure. When pysam is built to use the bundled HTSlib, it always runs HTSlib's _configure_ which does on these certain systems add the target flags in question to its generated _config.mk_:

```
$ python3 setup.py build
# pysam: cython is available - using cythonize if necessary
# pysam: htslib mode is shared
[…]
checking whether C compiler accepts -mssse3 -mpopcnt -msse4.1... yes
checking whether C compiler accepts -mavx2... yes
checking whether C compiler accepts -mavx512f... yes
checking whether C compiler supports ARM Neon... no

$ tail htslib/config.mk
[…]
# Extra CFLAGS for specific files
HTS_CFLAGS_AVX2 = -mavx2
HTS_CFLAGS_AVX512 = -mavx512f
HTS_CFLAGS_SSE4 = -mssse3 -mpopcnt -msse4.1
HTS_HAVE_NEON = 
```

However pysam does not use _config.mk_ or the _Makefile_'s rules directly to build these object files. Instead it queries `LIBHTS_OBJS` via `make print-config` and uses Python machinery to have each listed object file built using Python-appropriate flags.

Hence fixing this in due course will require

1. Probably having these `HTS_CFLAGS_*` variables printed out via HTSlib's `make print-config` so that they can be easily communicated to _setup.py_.
2. Figuring out how to add some target-specific options to particular object files when using the Python machinery.

jmarshall:
[...]

PR #1138, opened 2022-10-20
import htslib 1.16, bcftools 1.16 and samtools 1.16.1

AndreasHeger:
Updating to htslib-1.16, bcftools-1.16, samtools-1.16.1.

Notes:

The inclusion of the codecs caused an "invalid instruction" error at runtime. A solution that worked was to disable usage of all special instruction sets. It is implemented by the not very elegant method of removing the relevant DEFINE instructions in config.h.

@jmarshall , if you have a moment and some advice, I would very much appreciate it. My gut feeling is that for the embedded htslib it might be prudent to turn off `HAVE_SSE4_1`, `HAVE_SSSE3`, `HAVE_AVX2`, `HAVE_AVX512` to make sure pysam will run on a wide variety of systems. If there is a better way to do this than manipulating config.h, please let me know.

Longer term I think it might be best to remove htslib from within pysam altogether and link externally. This should work well with conda, but I am not sure if that will break everything for those installing from wheels.


jkbonfield:
I'm curious how the pysam code managed to execute a function without the appropriate instruction set.  Htscodecs has a cpuid check and dispatches to the correct implementation based on the system it is running.

Do you have a stack track showing which function it was in when it died?  Also details about the exact system it was running on (eg details gleaned from /proc/cpuinfo) so we can see better what happened.

I would be very wary about simply disabling this code, as it will make pysam unnecessarily slower for handling CRAM 3.1 format data.  Given 3.1 is still draft it's maybe not a major issue currently, but long term this isn't the right solution.

jmarshall:
I'm travelling at the moment, so reviewing this will be a little delayed.

I have a draft of this import in my **import-1.16** branch which I may have pushed to my fork. It preserved the workaround for the regidx duplication that we have used previously and has some other additions.

It also contains an in-progress conversion of the build process to use htslib's _Makefile_ to build _libhts.a_ and have the python stuff link against that. All that is a prerequisite for compiling an embedded libhts within pysam now that the htscodecs code requires all those build flags.

Building against an external libhts (with an external libhtscodecs) already works, as the Debian packagers have demonstrated (with pysam 0.19.1).

AndreasHeger:
Thanks @jkbonfield and @jmarshall for the quick response.

A stack trace and bits of cpuinfo are below. I think the issue arises because when pysam puts htslib together it does not go through the makefile meaning it uses a common set of compiler flags for every source file. I assume this is what causes the fail, as in the Makefile, the instruction set options are only set for the relevant htscodecs. I assume this is what @jmarshall  prompted to go through make to build libhts.

With your permisison, @jmarshall , I will see if I can find your import branch and take stuff from there - using the htslib makefile is the right solution I think.

Safe travels.

### stacktrace

These are the top bits of the stack-trace. There are an additional 40 levels all within python.
```
#0  0x0000155554d7c7d2 in kh_resize_vdict () from /home/andreas/priv/pysam/pysam/libchtslib.cpython-39-x86_64-linux-gnu.so
#1  0x0000155554d7f594 in kh_put_vdict () from /home/andreas/priv/pysam/pysam/libchtslib.cpython-39-x86_64-linux-gnu.so
#2  0x0000155554d81295 in bcf_hdr_add_hrec () from /home/andreas/priv/pysam/pysam/libchtslib.cpython-39-x86_64-linux-gnu.so
#3  0x0000155554d8c784 in bcf_hdr_parse () from /home/andreas/priv/pysam/pysam/libchtslib.cpython-39-x86_64-linux-gnu.so
#4  0x0000155554d8cb84 in vcf_hdr_read () from /home/andreas/priv/pysam/pysam/libchtslib.cpython-39-x86_64-linux-gnu.so
#5  0x0000155554d8d4c5 in bcf_hdr_read () from /home/andreas/priv/pysam/pysam/libchtslib.cpython-39-x86_64-linux-gnu.so
#6  0x0000155554545b2c in __pyx_pw_5pysam_7libcbcf_11VariantFile_15open () from /home/andreas/priv/pysam/pysam/libcbcf.cpython-39-x86_64-linux-gnu.so
#7  0x00005555556c8714 in cfunction_call (func=0x1555543cbb80, args=<optimized out>, kwargs=<optimized out>)
    at /tmp/build/80754af9/python-split_1631797238431/work/Objects/methodobject.c:543
#8  0x00001555544e67c6 in __pyx_pw_5pysam_7libcbcf_11VariantFile_3__init__ () from /home/andreas/priv/pysam/pysam/libcbcf.cpython-39-x86_64-linux-gnu.so
#9  0x00005555556989bb in type_call (kwds=0x0, args=0x155554e32c40, type=<optimized out>) at /tmp/build/80754af9/python-split_1631797238431/work/Objects/typeobject.c:1026
#10 _PyObject_MakeTpCall (tstate=0x555555914720, callable=0x15555459be80 <__pyx_type_5pysam_7libcbcf_VariantFile>, args=<optimized out>, nargs=<optimized out>, 
    keywords=<optimized out>) at /tmp/build/80754af9/python-split_1631797238431/work/Objects/call.c:191
#11 0x0000555555722d89 in _PyObject_VectorcallTstate (kwnames=0x0, nargsf=<optimized out>, args=0x155553e04e78, callable=<optimized out>, tstate=<optimized out>)
    at /tmp/build/80754af9/python-split_1631797238431/work/Include/cpython/abstract.h:116
#12 PyObject_Vectorcall () at /tmp/build/80754af9/python-split_1631797238431/work/Include/cpython/abstract.h:127
#13 call_function (kwnames=0x0, oparg=<optimized out>, pp_stack=<synthetic pointer>, tstate=0x555555914720)
    at /tmp/build/80754af9/python-split_1631797238431/work/Python/ceval.c:5075
#14 _PyEval_EvalFrameDefault (tstate=<optimized out>, f=0x155553e04cf0, throwflag=<optimized out>) at /tmp/build/80754af9/python-split_1631797238431/work/Python/ceval.c:3487
....
```
### cpuinfo
```
...
processor	: 11
vendor_id	: GenuineIntel
cpu family	: 6
model		: 158
model name	: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
stepping	: 10
microcode	: 0xf0
cpu MHz		: 3968.921
cache size	: 9216 KB
physical id	: 0
siblings	: 12
core id		: 5
cpu cores	: 6
apicid		: 11
initial apicid	: 11
fpu		: yes
fpu_exception	: yes
cpuid level	: 22
wp		: yes
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc art arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf pni pclmulqdq dtes64 monitor ds_cpl vmx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm 3dnowprefetch cpuid_fault epb invpcid_single pti ssbd ibrs ibpb stibp tpr_shadow vnmi flexpriority ept vpid ept_ad fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid mpx rdseed adx smap clflushopt intel_pt xsaveopt xsavec xgetbv1 xsaves dtherm ida arat pln pts hwp hwp_notify hwp_act_window hwp_epp md_clear flush_l1d arch_capabilities
bugs		: cpu_meltdown spectre_v1 spectre_v2 spec_store_bypass l1tf mds swapgs itlb_multihit srbds mmio_stale_data
bogomips	: 4399.99
clflush size	: 64
cache_alignment	: 64
address sizes	: 39 bits physical, 48 bits virtual
power management:
```

jmarshall:
That stack trace shows a crash nowhere near the codecs code. Looking at the now commented-out code added to _setup.py_, I think the crash is due to adding the AVX etc compile flags for **all** object files. This has led to the compiler using unavailable SIMD instructions while compiling the _htslib/vcf.c_ code.

As discussed in #1112, these compiler flags need to be used **only** for the particular _htscodes/rANS_static32x16pr_xyz.c_ code that they pertain to. The rest of the object files should be built just with their usual flags. Hence my proposed upcoming _setup.py_ changes to build _libhts.a_ using HTSlib's _Makefile_ and simply listing `extra_objects=["htslib/libhts.a"],` for all the modules in _setup.py_ (more or less).

jmarshall:
FYI my version of this on my **import-1.16.1** branch is now really pushed to my fork and is now PR #1140. You may wish to compare the best bits of the two.

I figured out what the pre-existing _setup.py_ buglet that was leaving `-fPIC` out of the command was, and corrected it. I hadn't seen that before because my local Linux box puts `-fPIC` into the command line two or three times (!) but the GItHub Actions Ubuntu boxes do not — as you found out the hard way in the last few commits here (thanks for leading the way on that bit!).

AndreasHeger:
Thanks. I will close this PR and go with your version.

jmarshall:
Excellent :smile: Note I just amended it (by force-pushing the PR branch) with a small improvement.

PR #1140, opened 2022-10-26
Update imports to htslib/samtools/bcftools 1.16[.1]

jmarshall:
Import htslib 1.16, samtools 1.16.1, and bcftools 1.16, and add the new `samtools reference` command to _pysam/samtools.py_.

Due to SIMD compiler flags needing to be used on only a handful of the HTSlib source files (as discussed in #1112), this one is a large change to the bundled htslib build process.

Instead of building the HTSlib source files via Python's extension building mechanisms, use HTSlib's _Makefile_ to build _libhts.a_, and have the Python extension modules link against this (effectively — in fact in the usual shared mode, they still link against the list of object files; see the comment in the code) instead of compiling the C source files themselves.

Note that Linux Python installations differ in whether `-fPIC` is listed in their Python `$CFLAGS`. There is also a fix to the flag probes in _setup.py_ so that `$CCSHARED` is added as appropriate.

Reorder `include_dirs` so that _pysam/version.h_ is picked up first.
Clean _libhts.a_ and `*.o` in the htslib/ subdirectory.
Ignore `*.o`, `libhts.a`, and `htslib/version.h`.
(These latter could be avoided by configuring and building HTSlib in a separate build directory under `build_temp` instead, but that would require more refactoring of _setup.py_.)


jmarshall:
Tested in `shared` mode. I haven't properly tested `separate` or `external` yet.

jmarshall:
Tested with `pytest` passing as usual in all modes: `shared`, `separate`,[^1] and `external`.

[^1]: When fef6567863e83c46557c5151e10fcaaf154f0eb7 has been applied, which it will have been if this is merged to **master**. D'oh!

AndreasHeger:
Thanks, I will merge and build the release notes.

I hope to have a bit of time in the next few days and there are three things I thought I should look at:
1. removing python 2.7 and legacy code
2. looking at the modified bases
3. see if AlignedSegment can be pickled

jmarshall:
Great news. I don't have any further PRs imminent at the moment that should go into a release.

This one last 2.7-supporting release, and then it would be great to remove all that legacy! A big :+1: to that.

I have been looking into the pickling thing a bit, and can share my notes.

AndreasHeger:
And also thanks for your work on the 1.16x upgrade.

PR #1142, opened 2022-10-28
No more py2

AndreasHeger:
Remove py2 support from pysam.


jmarshall:
That macOS `error: liblzma development files not found` build failure is something I've been meaning to look into for a long while for both htslib and pysam: recent macOS releases no longer include the same subset of lzma development files as previous releases. For some reason some dependencies that come along with previous Pythons paste over the cracks enough. For now, perhaps just leave Python 3.11 off…?

mdshw5:
@jmarshall I'm seeing the `liblzma` error message on macOS with both [python 3.10](https://github.com/mdshw5/pyfaidx/actions/runs/3630658557/jobs/6124278416#step:6:152) and [python 3.9](https://github.com/mdshw5/pyfaidx/actions/runs/3630658557/jobs/6124278752#step:6:152) as well. Do you have any idea about the best way to install `pysam` on the GitHub Actions `macos-latest` images?

jmarshall:
Make sure you have your own full liblzma installation as well as the system's partial one, e.g. via `brew install xz`.

mdshw5:
Thanks @jmarshall - that was my first thought, but the issue persists (even with `brew reinstall xz`): https://github.com/mdshw5/pyfaidx/actions/runs/3631904338/jobs/6127118303#step:3:1. It's not blocking me from releasing new packages as I only test against pysam, but it would still be nice to get macOS tests passing.

jmarshall:
That appears to have failed installing Python prerequisites, rather than building pysam itself. But I can't click through to the details on my phone 😄 

(Unfortunately I'm travelling at the moment and won't be able to look at this in depth for at least a couple of weeks.)

mdshw5:
No worries @jmarshall - I'll leave it off here, and will only tag you further if I identify a solution. Enjoy your travels!

jmarshall:
@mdshw5: I'm finally back looking into this problem, and of course by now your logs have expired. Do you have any more recent builds failing due to this or any other further information? No problem if not, I can do some test runs to try to reproduce the problem…

jmarshall:
@mdshw5: Now fixed by PR #1207. It turns out that for **multiarch** macOS builds (I suspect your builds were during a period in which the Pythons 3.9 and 3.10 you used were also multiarch), you need to make sure that it **doesn't** find a brewed xz!

PR #1147, opened 2022-11-10
Get shared object suffix from `EXT_SUFFIX`

sirno:
Hey,

I ran into an issue installing pysam with python 3.11.0. It seems that the `SO` configuration has been deprecated and `EXT_SUFFIX` should be used instead.

Also note that using `get_config_var` to access would have likely triggered a deprecation warning :)

Lets see what the actions say, but in case this causes test failures for older versions, I can adjust this to make it conditional.

Cheers

### DIFF
--- a/cy_build.py
+++ b/cy_build.py
@@ -8,7 +8,7 @@ except ImportError:
     from setuptools.command.build_ext import build_ext
 
 from distutils.extension import Extension
-from distutils.sysconfig import get_config_vars, get_python_version
+from distutils.sysconfig import get_config_var, get_config_vars, get_python_version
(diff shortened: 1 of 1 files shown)


mvdbeek (approved):
That works for me, on 3.11 and 3.6, so I guess the tests would pass.

jmarshall:
Thanks, I've merged this and updated it to use both `EXT_SUFFIX` and `SO` everywhere. (According to _pysam/NEWS_ we still support back to Python 3.2. Some early 3.2.x and 3.3.x versions do not define
EXT_SUFFIX, so for now we still need to consult both config variables.)

PR #1157, opened 2023-01-05
Fix type hints and add hints for FastqProxy

tfwillems:
This PR fixes type hits for get_tid in AlignmentHeader and for each of the fields in FastxRecord, which should be Optional[str] instead of str.

I also added some stubs for the FastqProxy class which were missing

Thanks for all your work on maintaining such a great library!


AndreasHeger:
Many thanks, apologies for the delay.

PR #1158, opened 2023-01-06
Fix type hint for IteratorRow's __next__

tfwillems:
Since IteratorRow is returned by fetch, and this should yield alignments instead of pileup columns, this return type should be AlignedSegment

Reverts change made in 548a5021cb015537add5f2d817d8d892f245f10c


AndreasHeger:
Thanks!

PR #1173, opened 2023-03-17
Fix building htscodecs SIMD code on multi-arch macOS

### LINKED ISSUE #1149: fails on import in m1 MacBook
I can't seem to import pysam after it appears to install successfully.

```
$ python -m vent testenv 

$ . testenv/bin/activate

$  pip install --upgrade pip
Requirement already satisfied: pip in ./testenv/lib/python3.10/site-packages (22.2.2)
Collecting pip
  Using cached pip-22.3.1-py3-none-any.whl (2.1 MB)
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 22.2.2
    Uninstalling pip-22.2.2:
      Successfully uninstalled pip-22.2.2
Successfully installed pip-22.3.1

$ pip install pysam 
Collecting pysam
  Using cached pysam-0.20.0.tar.gz (4.0 MB)
  Preparing metadata (setup.py) ... done
Using legacy 'setup.py install' for pysam, since package 'wheel' is not installed.
Installing collected packages: pysam
  Running setup.py install for pysam ... done
Successfully installed pysam-0.20.0
```

```
$ python
Python 3.10.7 (v3.10.7:6cc6b13308, Sep  5 2022, 14:02:52) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
>>> import pysam
  File "/Users/gh13047/testenv/lib/python3.10/site-packages/pysam/__init__.py", line 4, in <module>
    from pysam.libchtslib import *
ImportError: dlopen(/Users/gh13047/testenv/lib/python3.10/site-packages/pysam/libchtslib.cpython-310-darwin.so, 0x0002): symbol not found in flat namespace (_rans_compress_O0_32x16_neon)
```

This is only in M1 MacBook, I'm able to install it fine on a linux server with intel processor

jmarshall:
On macOS, hack the htslib (really just htscodecs) build a little bit so that the htscodecs SIMD code is only built for one architecture, i.e., either x86-64 or arm64 as appropriate. The next versions of htslib and htscodecs will likely use a different approach for this (cf samtools/htscodecs#76 and the resulting PRs), but applying this simple approach suffices in the context of pysam's more regimented build process and fixes #1149 now (and can be revisited when we import an htslib in which the issue is addressed upstream).

----

On recent macOS, universal builds are done with `-arch x86_64 -arch arm64`. This breaks configure tests for htscodecs's special SIMD implementations.

Use separate CFLAGS_ARM/_X86 variables to run these configure tests and build these particular translation units for only one architecture, and hack configure to set these to use only one -arch option as appropriate. Fixes #1149. This is something of a hack and upstream htscodecs will likely use a different approach, but will nonetheless be useful for pysam in the interim.


AndreasHeger:
Thanks, @jmarshall

AndreasHeger (approved):

Issue #1169, opened 2023-02-19
Test failures on musl with python 3.10


SoapGentoo:
I'm reporting this from Gentoo: https://bugs.gentoo.org/836857

I'm updating Gentoo's pysam to 0.20.0 and tested musl again. This is on musl 1.2.3 with python 3.10.10:

```
tests/AlignmentFileHeader_test.py::TestHeaderCRAM::test_dictionary_access_works Fatal Python error: Segmentation fault

Current thread 0x00007f1922ca4b48 (most recent call first):
  File "/var/tmp/portage/sci-biology/pysam-0.20.0/work/pysam-0.20.0-python3_10/tests/AlignmentFileHeader_test.py", line 197 in tearDown
  File "/usr/lib/python3.10/unittest/case.py", line 552 in _callTearDown
  File "/usr/lib/python3.10/unittest/case.py", line 594 in run
  File "/usr/lib/python3.10/unittest/case.py", line 650 in __call__
  File "/usr/lib/python3.10/site-packages/_pytest/unittest.py", line 330 in runtest
  File "/usr/lib/python3.10/site-packages/_pytest/runner.py", line 167 in pytest_runtest_call
  File "/usr/lib/python3.10/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/usr/lib/python3.10/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/usr/lib/python3.10/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/usr/lib/python3.10/site-packages/_pytest/runner.py", line 260 in <lambda>
  File "/usr/lib/python3.10/site-packages/_pytest/runner.py", line 339 in from_call
  File "/usr/lib/python3.10/site-packages/_pytest/runner.py", line 259 in call_runtest_hook
  File "/usr/lib/python3.10/site-packages/_pytest/runner.py", line 220 in call_and_report
  File "/usr/lib/python3.10/site-packages/_pytest/runner.py", line 131 in runtestprotocol
  File "/usr/lib/python3.10/site-packages/_pytest/runner.py", line 112 in pytest_runtest_protocol
  File "/usr/lib/python3.10/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/usr/lib/python3.10/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/usr/lib/python3.10/site-packages/pluggy/_hooks.py", line 265 in __call__
[...]

PR #1178, opened 2023-03-27
updated release notes

AndreasHeger:


AndreasHeger:
Thanks, @jmarshall , if you want to have a quick look at the release notes, please feel free to do so.

Special thanks for your work on the osX build will go into the github release notes!

NEWS
```diff
@@ -5,18 +5,22 @@ http://pysam.readthedocs.io/en/latest/release.html
...
+support. Pysam is tested with python versions 3.7 to 3.11.
+
+* [#1175] VariantHeader.new_record: set start/stop before alleles
+* [#1173] Add multiple build improvements in htscodecs on multi-arch macOS
+* [#1148] Ignore CIGAR-less reads in find_introns.
+* [#1172] Add new `samtools cram-size` and `samtools reset` commands
+* [#1168] Add a minimal pyproject.toml for PEP517.
+* [#1158] Fix type hints and add FastqProxy type hints.
+* [#1147] Py3.11 compatibility, get shared object suffix from EXT_SUFFIX.
+* [#1143] Add mypy symbols for samtools and bcftools.
```
jmarshall, line 23: These will be largely covered by the build improvements bit you mentioned you'll add to the GitHub release message, but as there are issue numbers for them that people might look for, we might add brief notes to NEWS too:

```
[#1155] Fix pysam.index() when using recent `samtools index` options.
[#1151] Test suite py3.11 compatibility, work around failing test case.
[#1149] MacOS universal build compatibility.
[#1146] Fix build when CFLAGS/etc environment variables are set.
```

jmarshall (commented):
Looks good, a couple of suggestions.

jmarshall:
> Pysam is now compatible with 3.11. We have removed python 2.x support. Pysam is tested with python versions 3.7 to 3.11.

It'd be good to have some clarity about exactly which Python 3.x versions we aim to support.

Currently (removed by this PR) _NEWS_ suggests the minimum is 3.2. _pyproject.toml_ says 3.8 (I hope this won't cause any trouble for people using 3.6 or 3.7). Our CI tests check 3.7 through 3.11, and only dropped 3.6 because GitHub's Ubuntu images no longer contain it.

I suspect many current extant HPC deployments' system Python installation is still 3.6 — e.g. various CentOS or Ubuntu installations.

So ideally I think we would say we support Py 3.6 or later. (Or stick with the existing 3.2 or later, but there's a few things missing from those old versions that's a bit painful.)

NEWS
AndreasHeger: Thanks, added!

AndreasHeger:
That makes sense. I will do a quick check which python versions still work.

jmarshall:
I fixed one more bug (#1169) on **master**, so have added a NEWS entry for it.

AndreasHeger:
Thanks. I tested under 3.6, but failed getting any older python version to install and did not have time to setup separate test environments for older pythons.

Setting minimum python version to 3.6 and saying tested on 3.6 - 3.11.

PR #1186, opened 2023-05-03
Remove cython from runtime dependencies

nsoranzo:
It's only needed at build time.

Should fix https://github.com/galaxyproject/galaxy/pull/15942/files#r1167475501

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,11 +13,6 @@ dynamic = [
     "readme",
 ]
 
-dependencies = [
-    "cython",
-]
-
-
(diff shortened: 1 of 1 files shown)


nsoranzo:
Ping @jmarshall @AndreasHeger , this affects also some bioconda recipes linked above.

jmarshall:
In order to benefit from this fix, there would also need to be a new point release made. This is @AndreasHeger's department, so there is no need to ping me.

If this is causing trouble for bioconda, they have the option of patching their package themselves.

Issue #1184, opened 2023-04-14
Iterating over AlignmentFile not properly type-hinted


PikalaxALT:
In `libcalignmentfile.pyi`, the return types for `__iter__` and `__next__` are given as `Any` when they should be `AlignmentFile` and `AlignedSegment` respectively. The current situation impedes type checking by linters.

jmarshall:
PR #1188 presents two approaches for addressing this.

Issue #1053, opened 2021-10-22
Unexpected results when unpacking `pysam.libctabixproxies.TupleProxy` during fetching a BED file


mrjng:
Hi, pysam developers.

There were unexpected results when I assigned the fetched result to the starred expression.
Here are more details.

### Situation
**Test file**
_test.bed.gz_ (compressed using _bgzip_ and indexed using _tabix_)
```
chr1    100     200     1
chr1    300     400     1
chr1    500     600     1
chr1    700     800     1
```

**Test script**
```python
import pysam

test_bed_path = 'test.bed.gz'
output_path = 'output.bed'

with pysam.TabixFile(test_bed_path) as bed_file, \
    open(output_path, 'w') as outfile:
    for fields in bed_file.fetch('chr1', parser=pysam.asTuple()):
        config, start, end, *others = fields
        print(config, start, end, *others, sep='\t', file=outfile)
```

**Expected output**
_output.bed_ will be the same with _test.bed.gz_

**Actual output**
```
chr1    100     200     chr1    100     200     1
chr1    300     400     chr1    300     400     1
chr1    500     600     chr1    500     600     1
chr1    700     800     chr1    700     800     1
```

In the test script, _others_ contained all the fields of the input bed file (_test.bed.gz_). (e.g. ['chr1', '100', '200', '1']...)

On the other hand, if the fetching results (`pysam.libctabixproxies.TupleProxy`) were cast into `tuple`s, the results were what I expected. Here is the other test script for this statement.
```python
import pysam

test_bed_path = 'test.bed.gz'
output_path = 'output.bed'

with pysam.TabixFile(test_bed_path) as bed_file, \
    open(output_path, 'w') as outfile:
    for fields in bed_file.fetch('chr1', parser=pysam.asTuple()):
        config, start, end, *others = tuple(fields)
        print(config, start, end, *others, sep='\t', file=outfile)
```

There might be a bug in `pysam.libctabixproxies.TupleProxy`. Please take a look at this issue.
Thanks.

PR #1188, opened 2023-05-08
Fix AlignmentFile iteration type hints by adding AlignmentFileIterator class

jmarshall:
Issue #1184 notes that iteration over `AlignmentFile` is currently not type-hinted:

```python
f = pysam.AlignmentFile('foo.bam')
for rd in f:
    …
```

We could fix the type hints directly, but at present the whole-file iteration class is `AlignmentFile` itself and it would be preferable not to expose that implementation detail. One day for example we may need to do something like the fix for #1053, which involved introducing a new class to manage iteration for `TupleProxy`.

So this PR creates a separate class for that now to avoid this problem: Move whole-file iteration to a special-purpose class instead of doing it within `AlignmentFile` itself. Add `__iter__`/`__next__` type annotations accordingly, so that mypy can determine the type of `rd` (as `AlignedSegment`) in loops like the one above.

----

An alternative would be to use a type alias to signal the potential difference from `AlignmentFile`, e.g.,

```diff
--- a/pysam/libcalignmentfile.pyi
+++ b/pysam/libcalignmentfile.pyi
@@ -71,6 +71,10 @@ class AlignmentHeader:
     def is_valid_tid(self, tid: int) -> bool: ...
     def get_tid(self, reference: str) -> int: ...
 
+# The iterator produced by AlignmentFile is currently itself, but this may
+# change in future and code should not make assumptions about this type.
+AlignmentFileIterator = AlignmentFile
+
 class AlignmentFile(HTSFile):
     def __init__(
         self,
@@ -172,8 +176,8 @@ class AlignmentFile(HTSFile):
     def nocoordinate(self) -> int: ...
     def get_index_statistics(self) -> List[IndexStats]: ...
-    def __iter__(self) -> Any: ...
-    def __next__(self) -> Any: ...
+    def __iter__(self) -> AlignmentFileIterator: ...
+    def __next__(self) -> AlignedSegment: ...
     def is_valid_tid(self, tid: int) -> bool: ...
     def get_tid(self, reference: str) -> int: ...
```

However I don't think the compatibility issues of making `AlignmentFileIterator` a real type are too great, so IMHO the PR change is better.


jmarshall:
The second commit shows the compatibility issue of needing to use `iter()` explicitly to turn an `AlignmentFile` into an iterator. No changes are needed for a loop:

```python
for rd in fp:
    …process rd
```

but if `next()` is used, you need to ask for the iterator explicitly:

```python
fp = AlignmentFile("foo.bam")
rd1 = next(fp)
rd2 = next(fp)
```

needs to be something like

```python
fp = AlignmentFile("foo.bam")
fpiter = iter(fp)
rd1 = next(fpiter)
rd2 = next(fpiter)
```

If code like the latter with explicit `next()` is common, I guess that would be a reason to use the simpler type alias alternative instead.

jmarshall:
On second thoughts: I expected there was an `AlignmentFile.read()` call or similar, but it seems that iteration is the main way of ad-hoc reading a SAM/BAM/CRAM file sequentially. Hence the `next()` code probably is quite common, so the compatibility cost of introducing `AlignmentFileIterator` and needing to add `iter(…)` all over the place — even though doing so would be arguably correct — is probably too high.

Hence I think it would probably be better to apply PR #1189 (which is the type alias alternative mentioned above) instead of this one. But I'll leave both here for other maintainers' consideration.

PR #1198, opened 2023-06-23
samtools/*.c: fix a couple of typos.

emollier:
This patch addresses a couple of typos caught by lintian while working on the packaging of pysam for Debian.


jmarshall:
These files are imported from samtools. So these typo fixes would need to be sent upstream to the [samtools repository](https://github.com/samtools/samtools).

emollier:
Ah, yes, thanks for the reminder, that also explains me why things looked familiar when I came accross the lintian flags in samtool.  (:

Going ahead...

PR #1212, opened 2023-08-21
Add HTSFile.seek(..., whence) parameter & clarify libc. vs io.SEEK_SET

### LINKED ISSUE #1185: HTSFile.seek(whence)
Can you add a "whence" parameter to the HTSFile.seek()?

jmarshall:
[`DEF` is deprecated in Cython 3](https://cython.readthedocs.io/en/latest/src/changes.html#deprecations), so we revisit the `SEEK_*` definitions in _libchtslib.pyx_. In fact pysam should not be defining these at all:

* low-level functions directly from HTSlib (such as bgzf_seek()) take the OS's SEEK_SET/etc as defined in libc.stdio / posix.unistd / etc

* pysam's exported Python functionality should really take a parameter type that is available to Python programs, namely io.SEEK_SET etc

Hence remove the DEFs and clarify which functions use libc.SEEK_* and which use io.SEEK_*. Convert between the two where necessary.

While we're here, add an optional `whence` parameter to HTSFile.seek(). Note that some of the underlying APIs may not accept any value other than SEEK_SET. Fixes #1185.

Also fix buglet in HFile.close(): hclose() reports the type of any error encountered in errno, as there's no longer an hFILE to call herrno() on (moreover here self.fp is already NULL).


bioinformed (approved):
LGTM -- thanks for the continued proactive maintenance!

PR #1213, opened 2023-08-25
Prepare for Python 3.12 by avoiding distutils in setup.py

jmarshall:
Python 3.12 has removed distutils, so — even though setuptools will provide an interim distutils shim — it's best to avoid it.

Incorporate _cy_build.py_ directly into _setup.py_, which facilitates moving `check_ext_symbol_conflicts()` to our `build_ext` subclass and avoid subclassing `build` at all (as this is only supported in very recent setuptools).

Import various items from setuptools instead of distutils.


bioinformed (approved):
As one of the original authors of cy_build.py (along with Josh Paul), I'm shocked it survived this long.  Thanks for your continued support of pysam.

jmarshall:
Thanks for reviewing this.

The _cy_build.py_ code that was still being used is still there, but I'm planning to re-examine it after the upcoming release. 😄

PR #1216, opened 2023-09-08
Move Cython to build dependencies in `setup.py`

althonos:
Hi!

This PR moves Cython from the `requires` key to the `setup_requires` key in `setup.py`. With the current requirement, `pysam` would require Cython to be installed on the target machine, even when installing from a wheel, where no Cython compilation is necessary. By moving Cython to the `setup_requires` field, it indicates Cython is needed to build the package, but not actually needed to import the package once it has been built.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -684,7 +684,7 @@ metadata = {
     'classifiers': [_f for _f in classifiers.split("\n") if _f],
     'url': "https://github.com/pysam-developers/pysam",
     'packages': package_list,
-    'requires': ['cython (>=0.29.12)'],
+    'setup_requires': ['cython (>=0.29.12)'],
     'ext_modules': [CyExtension(**opts) for opts in modules],
     'cmdclass': {'build_ext': cy_build_ext, 'clean_ext': clean_ext, 'sdist': cythonize_sdist},
     'package_dir': package_dirs,


jmarshall:
The `requires` entry does not cause Cython to be installed even when installing from a wheel — that would happen if we used `install_requires`, which we don't. There was however a bug in 0.21.0's _pyproject.yaml_ that caused Cython to be unnecessarily installed, which has since been fixed: see #1186.

[`setup_requires` is discouraged in favour of _pyproject.toml_](https://setuptools.pypa.io/en/stable/references/keywords.html#keyword-setup-requires), which we already use. So it would be better to remove this entry than to change it to `setup_requires`.

In fact, it doesn't really seem that this `requires` entry actually does anything any more anyway! It adds a `Requires: cython (>=0.29.12)` line to the resulting wheel's _pysam-x.y.z.dist-info/METADATA_ file, but that's a v1.1 item that was [superseded in 2005](https://peps.python.org/pep-0345/#summary-of-differences-from-pep-314) and I suspect nothing actually acts on it now.

So it would be best to remove this line. If we do that, it would be best to update the setuptools requirement to v61.0.0, which is documented as the version that introduced robust _pyproject.toml_ support. But I am in two minds whether to do that now or to wait until after the upcoming release…

jmarshall:
Thanks for bringing this up. I've incorporated something based on your commit into PR #1219, so we can review these related changes together.

althonos:
Okay thanks. I was just a bit surprised because a `pip install` of `pysam` would also install Cython, even when installing `pysam` from a wheel, so I figured there was something wrong with the dependencies:

image

Issue #1217, opened 2023-09-08
more stringent cython version pin


ahvigil:
The semi-recent release of cython 3 has done a number on a lot of my projects which pin older versions of pysam, due to various incompatibilities introduced with the major cython release. Furthermore, build isolation features in newer versions of pip mean its not enough to force a cython version before installing pysam to get it to install correctly, so a fairly common workflow for me nowadays is installing pysam with `--no-build-isolation` so I can manually install `Cython<3`

I think it might be a good idea for future releases of pysam to specify both lower _and_ upper bounds on the required cython version. For example, the current requirement should be `Cython>=0.29.12,<4` so that today's releases aren't rendered broken by future breaking changes in cython. Otherwise isolated pysam builds always try to use the newest cython.

jmarshall:
Pysam was reasonably quick to support Cython 3, and the issues for older pysam versions were only minor. Nonetheless you are right that this would be worth doing, as long as it's made clear that there is no intrinsic bar on an eventual Cython 4 and the pysam of the day will be updating to be compatible with it.

However we go to some lengths to pre-cythonise our sdists, so for people installing via pip, Cython should not actually be necessary. The circumstances in which it would be necessary are when installing pysam for a new Python release that the Cython in use back when the pysam release was made did not support. In this case you would want to regenerate the _*.c_ files with current Cython.

So ideally perhaps we would remove Cython from _pyproject.yaml_, while leaving it in _requirements-dev.txt_, which can be used by those (including readthedocs) building from a git repository, which does not contain the _*.c_ files. But that might not be practical, as we would need to have Cython as an optional build requirement (which I am not sure is a thing); or perhaps the maintainers could regenerate existing release's sdists with new Cythons when necessary.

TL;DR we should add `,<4` in _pyproject.toml_, probably in _requirements-dev.txt_, but not in _setup.py_ (cf https://github.com/pysam-developers/pysam/pull/1216#issuecomment-1712498809).

ahvigil:
> However we go to some lengths to pre-cythonise our sdists, so for people installing via pip, Cython should not actually be necessary. The circumstances in which it would be necessary are when installing pysam for a new Python release that the Cython in use back when the pysam release was made did not support. In this case you would want to regenerate the *.c files with current Cython.

[...]

PR #1218, opened 2023-09-08
Pin cython<4

ahvigil:
addresses https://github.com/pysam-developers/pysam/issues/1217 by adding a pin to cython <4 so future cython releases don't break current version builds.


jmarshall:
Thanks for bringing this up. I've incorporated your commit into PR #1219, so we can review these related changes together.

jmarshall:
Thanks. I've now merged a squashed version of this as a23ba53a4db9edabafeb9aa5a0a96b9c5ee6ae43.

Issue #1237, opened 2023-10-18
AlignmentHeader.from_dict() silently discards the TP attribute from SQ lines


tfenne:
The `TP` attribute in sequence lines in the SAM header is used to encode whether a reference sequence is linear or circular.  Unfortunately `AlignmentHeader.from_dict()` silently removes this tag from the data:

```
> header = { "SQ": [{"SN": "some-plasmid", "LN": 5000, "TP": "circular"}] }
> AlignmentHeader.from_dict(header).to_dict()
OrderedDict([('SQ', [{'SN': 'some-plasmid', 'LN': 5000}])])
```

jmarshall:
I really dislike this “I only believe in fields I know about” aspect of this part of pysam and would rather remove it. Those tables of known attributes are useful to return `LN` as `int` and so on, but IMHO unknown-at-the-time attributes should be preserved and default to `str` instead of being dropped.

Perhaps that's a quick fix too, but in the meantime PR #1238 would be a good stopgap.

tfenne:
I totally agree @jmarshall.

Quoting the spec on the header (emphasis in the second paragraph mine):

> 1.3 The header section
> Each header line begins with the character ‘@’ followed by one of the two-letter header record type codes
defined in this section. In the header, each line is TAB-delimited and, apart from @CO lines, each data field
follows a format ‘TAG:VALUE’ where TAG is a two-character string that defines the format and content of
VALUE. Thus header lines match `/^@(HD|SQ|RG|PG)(\t[A-Za-z][A-Za-z0-9]:[ -~]+)+$/` or `/^@CO\t.*/`.  Within each (non-@CO) header line, no field tag may appear more than once and the order in which the fields
appear is not significant.
>
> The following table describes the header record types that may be used and their predefined tags. Tags
listed with ‘*’ are required; e.g., every @SQ header line must have SN and LN fields. As with alignment optional
fields (see Section 1.5), **you can freely add new tags for further data fields**. Tags containing lowercase letters
are reserved for local use and will not be formally defined in any future version of this specification.

jmarshall:
Yes, I remember writing that text :smile: — though it was longer ago than I expected, in 2015.

It turns out that `KNOWN_HEADER_FIELDS` as used by `to_dict()` **does** default to `str`. However `from_dict()` uses `VALID_HEADER_ORDER` to constrain which non-lowercase fields it will process. So I've merged PR #1238 to improve this in the short term.

[...]

PR #1238, opened 2023-10-18
Add missing SQ tags as per the spec

nh13:
See #1237


pysam/libcalignmentfile.pyx
```diff
@@ -127,7 +128,8 @@ KNOWN_HEADER_FIELDS = {"HD" : {"VN" : str, "SO" : str, "GO" : str},
 # the end as parsing a CL will ignore any subsequent records.
 VALID_HEADER_ORDER = {"HD" : ("VN", "SO", "GO"),
                       "SQ" : ("SN", "LN", "AS", "M5",
-                               "UR", "SP", "AH"),
+                              "UR", "SP", "AH", "TP",
+			      "DS", "AN"),
```
nh13, line 132: These are not in the same order as the spec, so we could take the time to update them?
Screenshot 2023-10-18 at 4 53 45 PM

jmarshall (approved):

jmarshall:
Thanks. I've added a couple missing from other headers too. No need to reorder them just now, as I think the way forward for this functionality is for these two tables to largely go away — see further comments on #1237.

tfenne:
Thanks @jmarshall !

PR #1245, opened 2023-11-25
Add FastxRecord to the docs

nh13:


jmarshall:
When we add a previously omitted class to _api.rst_, we suddenly find out whether its docstrings render well as reStructuredText or provide useful information at all. In this case, some of it's not great:

> `name`
>  &nbsp;&nbsp;&nbsp;&nbsp;unicode
>  &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;**Type:** name

These appear to be the default rendering for `cdef public str` from the _.pxd_ file. (These are the only `public str` items in pysam.) “Unicode” is irrelevant as we're Python 3 only, and “Type” is just wrong.


Perhaps we can add some docstrings to the _.pyx_ file to override this rendering and provide some meaningful information for these properties…

jmarshall (approved):

nh13:
Thanks @jmarshall!

PR #1250, opened 2023-11-30
htslib: Fix mmap configure check for current compilers

fweimer-rh:
It is necessary to define _DEFAULT_SOURCE along with _XOPEN_SOURCE, otherwise the configure check for mmap fails with compilers which do not support implicit function declarations because the check relies on the presence of the getpagesize function.

Downstream version of samtools/htslib#1711.

### DIFF
--- a/htslib/config.h.in
+++ b/htslib/config.h.in
@@ -147,6 +147,9 @@
 # define _DARWIN_USE_64_BIT_INODE 1
 #endif
 
+/* Enable default extensions. */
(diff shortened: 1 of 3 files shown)


jmarshall:
Thanks for caring enough to provide a PR for pysam. However this is an HTSlib problem that I anticipate may be resolved in a different way, so we'll close this here and revisit when HTSlib has responded to the problem.

PR #1267, opened 2024-02-13
Add AlignedSegment.__repr__

marcelm:
Before:
```
<pysam.libcalignedsegment.AlignedSegment object at 0x7fd74c3ae2c0>
```
After:
```
<AlignedSegment('name', flags=16, ref='chr1', pos=1170070, mapq=255, cigar='50M', ...)>
```

I’ve used pysam with this patch for a while and find it very valuable when looking at `AlignedSegment`s in a debugger or the REPL.

I’ve chosen to include only the (from my perspective) most important attributes so that the output doesn’t become too long, in particular when there’s a list with `AlignedSegment`s.

The angle brackets are supposed to convey that the string cannot be passed to `eval()` to recreate the object. I’d be fine with removing them.

### DIFF
--- a/pysam/libcalignedsegment.pyx
+++ b/pysam/libcalignedsegment.pyx
@@ -978,6 +978,9 @@ cdef class AlignedSegment:
                                    self.query_qualities,
                                    self.tags)))
 
+    def __repr__(self):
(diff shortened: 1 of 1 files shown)


marcelm:
Great, thanks!

PR #1270, opened 2024-02-26
Automatically translate VariantRecord on write

kkchau:
Enable auto-translation of VariantRecord headers on write when a different header is detected.

Current functionality does not check if headers match on write, potentially resulting in (silent and dangerous) reshuffling of fields, e.g. when header records are ordered differently. Uncomment the lines that detects header differences and translates the VariantRecord to the new header to be written.

### DIFF
--- a/pysam/libcbcf.pyx
+++ b/pysam/libcbcf.pyx
@@ -3033,6 +3033,7 @@ cdef class VariantRecord(object):
         return makeVariantRecord(self.header, bcf_dup(self.ptr))
 
     def translate(self, VariantHeader dst_header):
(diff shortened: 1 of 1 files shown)


kkchau:
These commented lines were added in https://github.com/kkchau/pysam/commit/c30ead4825d588fe9d06d44f0028bf86ee6c6882 and https://github.com/kkchau/pysam/commit/36eb1dae8e21d9b031635f9df5fccf51c80f7642 but commented out. Not sure if there is still work that needs to be done (if so, I can close this PR), but seems to work after a couple of tests

PR #1273, opened 2024-03-16
fix: add missing properties to AlignedSegment type stubs

msto:
https://github.com/pysam-developers/pysam/pull/1088 added `is_mapped` and `is_forward` (and their mate equivalents) to the `AlignedSegment` API, but didn't declare these properties in the corresponding type stubs. Their absence here causes downstream woes with mypy's type checking


jmarshall (approved):

msto:
Thanks @jmarshall !

Issue #958, opened 2020-09-30
bcftools annotate method missing


tirohia:
Currently assuming this is the only method missing. Most methods for bcftools are present, I can't for the life of me, figure out how to call bcftools annotate though. It results in:

```
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/home/bcurran/.pycharm_helpers/pydev/_pydev_bundle/pydev_umd.py", line 197, in runfile
    pydev_imports.execfile(filename, global_vars, local_vars)  # execute the script
  File "/home/bcurran/.pycharm_helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/red/project/prosper/src/vcftojson.py", line 16, in <module>
    pysam.annotate("-a", "geneNames.bed.gz", "-c", "CHROM, FROM, TO, GENE", "-h", '##INFO=<ID=GENE,Number=1,Type=String,Description="Gene name">', "-o", outfile)
AttributeError: module 'pysam' has no attribute 'annotate'
```

Have also looked using dir, i.e.
```
$ python
Python 3.6.12 |Anaconda, Inc.| (default, Sep  8 2020, 23:10:56) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pysam
>>> dir(pysam)
[...]

PR #1275, opened 2024-03-31
+ document how to run bcftools commands from pysam

indraniel:
I recently had trouble running the `bcftools` commands within pysam.  I saw the ["Using samtools commands within python" ][0] section within the documentation and thought things would behave similarly. I didn't find it to be true.  I ended up finding the solution through [github issue 958][1].  

I thought it would be nice to have this `bcftools` feature more explicitly documented for future users.

Let me know if any additional changes are desired.

[0]: https://pysam.readthedocs.io/en/stable/usage.html#using-samtools-commands-within-python
[1]: https://github.com/pysam-developers/pysam/issues/958

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -172,6 +172,25 @@ Note that this means that output from commands which produce output on
 stdout will not be available. The only solution is to run samtools
 commands through subprocess.
 
+=====================================
+Using bcftools commands within python
+=====================================
+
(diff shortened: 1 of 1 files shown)


jmarshall:
Thanks for the suggestion to document the bcftools subcommands too. I've taken an alternative approach in 51ff3f1f1df3a452b43afcc5b122ff7510b70aa5 and incorporated bcftools into the existing samtools documentation.

PR #1295, opened 2024-06-25
Improve cigarstring performance

limwz01:
Use direct string manipulation instead of inefficient Python joins

### DIFF
--- a/pysam/libcalignedsegment.pyx
+++ b/pysam/libcalignedsegment.pyx
@@ -1316,10 +1316,11 @@ cdef class AlignedSegment:
                 else:
                     break
 
-            ret = force_str(buf.s[:buf.l])
-
-            if buf.m:
-                free(buf.s)
+            try:
+                return buf.s[:buf.l].decode("utf8")
+            finally:
+                if buf.m:
+                    free(buf.s)
 
             return ret
 
--- a/pysam/libcalignedsegment.pyx
+++ b/pysam/libcalignedsegment.pyx
@@ -1298,8 +1298,9 @@ cdef class AlignedSegment:
             cdef int ret
             cdef int pos = 0
             cdef int size = 16
-            cdef char buf[16]
-            cdef char *s = buf
+            cdef kstring_t buf
+            buf.l = buf.m = 0
+            buf.s = NULL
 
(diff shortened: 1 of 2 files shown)


jmarshall:
This PR is certainly on my shortlist to merge in due course, though I'd probably prefer if it used `kstring` rather than manual memory management. Have you done any timing testing?

limwz01:
I did timing testing quite long ago, and there was substantial improvement, but I've just tested with the latest version on Python 3.12.1 using a ~100MB BAM file:

```python3
    tm = timeit.default_timer()
    with pysam.AlignmentFile(args.bam) as in_bam_f:
        for read in in_bam_f:
            read.cigarstring
    print("finished reading BAM in %f seconds"%(timeit.default_timer() - tm))
```

```
without optimisation: 5.821680s
with optimisation: 3.519799s
```

limwz01:
nice, kstring is even faster:

```
without optimisation: 5.821680s
with optimisation: 2.439370s
```

There was a problem using the pattern in to_string:

```python3
            ret = force_str(buf.s[:buf.l])

            if buf.m:
                free(buf.s)

            return ret
```

I had to revert to using try:

```python3
            try:
                return buf.s[:buf.l].decode("utf8")
            finally:
                if buf.m:
                    free(buf.s)
```

If anyone can help to make the original pattern work, that would be great!

jmarshall:
The TypeError was because there was still a `cdef int ret` declaration left over from your manual allocation code, so assigning a string to `ret` didn't work. But it's fine, now that we have dropped Python 2.x support I prefer to use `.decode(…)` directly anyway.

Thanks for the timing tests. The kstring implementation is faster because it reallocates the array and continues, whereas it looks like your original implementation allocated a new array and restarted, rather than copying the progress so far into the new buffer.

Thanks for this PR. I've tidied up some leftovers from your original implementation, and am now happy with an implementation that makes the cigarstring directly from the internal representation rather than by constructing a cigartuple first.

jmarshall (approved):

limwz01:
oh yes, thanks for cleaning up. My bad, I totally didn't remember and didn't remove the restarting process when I converted to use kstring.

PR #1296, opened 2024-07-06
fix(libcalignmentfile.pyi): update IteratorColumn with missing import + correct _iter_ object

mshunjan:
According to the v0.22.1 docs, you can iterate over data in a SAM file  using the pileup method, which should return an IteratorColumn PileupColumn. However, it instead returns an Iterator AlignedSegment. This is due to a missing import for the PileupColumn and an incorrect __iter__ call.


jmarshall (approved):

jmarshall:
Thanks — merged as 453acaa2dd2e09b21793633198ba72abdcbe5e84.

PR #1315, opened 2024-10-22
Fix typechecking for top-level samtools commands like `pysam.view(…)`

jmarshall:
As reported in PR #1306, mypy rejects `pysam.view(…)` et al because it doesn't know that they exist. This is due to the definition of `__all__` in _pysam/samtools.py_: either `HAVE_FINAL` is True for mypy so it skips the `__all__` definition, or the definition is complex enough that mypy does not evaluate it, or both (as #1306 did not fix the problem).

The first commit fixes this by lifting the `__all__` definition to the top-level and making it a simple assignment of a literal list. The second commit simplifies the rest of the code in _pysam/samtools.py_ and _pysam/bcftools.py_.

### DIFF
--- a/pysam/bcftools.py
+++ b/pysam/bcftools.py
@@ -1,65 +1,32 @@
-try:
-    from typing import Final
-    HAVE_FINAL = True
-except ImportError:
-    HAVE_FINAL = False
+import pysam.utils
 
-from pysam.utils import PysamDispatcher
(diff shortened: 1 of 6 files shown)


jmarshall:
This also removes the `pysam.samtools.import_(…)` alias of `pysam.samtools.fqimport(…)`, which was quietly added without review during PR #1143.

PR #1324, opened 2025-01-13
Support setting query_qualities using a string

nh13:


jmarshall:
I rather like this, although I am slightly in two minds whether it wants a separate entry point or overloading the parameter type is a good idea — the latter will complicate the type hints a bit.

Amusingly this is already possible via the deprecated `qqual` property…

nh13:
@jmarshall thanks for considering this, I'll let you know how you'd like to precede

nh13:
@jmarshall let me know what you think, we could use the [overload decorator in typing](https://docs.python.org/3/library/typing.html#overload) too?

jmarshall:
I'm working on extending this a little. My main concern is that having set it via a string, there's no (non-deprecated) way to get a string back out again — reading `rec.query_qualities` will return it as an array.

So I'm drafting adding another `query_qualities_str` property that would get/set this as a string. And `query_qualities` as per your PR would accept setting from either an array or a string, but return only an array as at present. Possibly also a `query_qualities_array` as well for explicitness, but that one's a bit over the top.

There's also a bunch of small bugs in this area which I'm squashing… 😞

nh13:
Thanks @jmarshall!  It would be nice not to have to use `qualitystring_to_array`, `array_to_qualitystring`, and `qualities_to_qualitystring` anymore.

PR #1334, opened 2025-03-22
Fix invalid-type message for creating header from dictionary

gtamazian:
The message for the type error when creating a SAM header from a dictionary misses type names. The following MWE causes the error by passing a dictionary instead of a list of dictionaries under the **PG** tag in the SAM header:
```
import pysam

header = {"PG": {"ID": "my-tool"}}
output = pysam.AlignmentFile("test.sam", "w", header=header)
```
The current invalid-type error message:
```
ValueError: invalid type for record %s: %s, expected %s
```
The fixed error message:
```
ValueError: invalid type for record PG: <class 'dict'>, expected <class 'collections.abc.Sequence'>
```


jmarshall:
Thanks. For new code we're mostly switching to f-strings for formatting, but this is the easiest fix for this thinko.

Issue #121, opened 2015-06-09
Improve performance of accessing AlignedSegment.qual


marcelm:
Currently, when accessing the `.qual` attribute of an AlignedSegment, the full sequence of quality values is computed every time. This results in unnecessarily bad performance especially when accessing `read.qual[i]` for more than one i. (See also the report on the mailing list from June 1).

AndreasHeger:
Added caching for the fields `.query_qualities`, `.query_alignment_qualities`, `.query_sequence` and `.query_alignment_sequence`. This should also speed-up access to `.qual`.

A potential problem is that the returned arrays are mutable, but the changes are currently not written to disk. The returned strings don't have that problem.

How I wish I had stayed away from properties and used only getters/setters.

marcelm:
Cool, thanks!

PR #1341, opened 2025-05-15
Add AlignedSegment.query_qualities_str and fix sequence and qualities caching

jmarshall:
On PR #1324 @nh13 requested being able to set `AlignedSegment.query_qualities` via a SAM/FASTQ-style base quality string. This expands that to be able to retrieve QUAL as a string too.

Add a `query_qualities_str` property implemented directly against the underlying `bam1_t` data structure rather than translating to an array first via `qualitystring_to_array()`/`array_to_qualitystring()`. Recode `query_qualities` to work more obviously directly against the `bam1_t` too, and fix several bugs in its caching layer. (See #121 for why this caching is important.) Enable **setting** `query_qualities` to use a string too, by delegating to the new property.

Fix some similar caching bugs in `query_sequence` too.


nh13:
Thank-you!

PR #1342, opened 2025-05-22
Fix type hints for get_aligned_pairs

gshiba:


jmarshall:
Good point, the existing type hint is insufficient.

We still support Python 3.9 unfortunately, so these will need to use `Optional[str]` etc instead of `str | None`.

gshiba:
Thanks for the tip -- switched to use `Optional`

jmarshall:
Thank you, this was very helpful. I've made some updates based on fairly exhaustive testing with mypy: in particular the `int` members of the tuple are sometimes Optional too, and the cigar field is returned as an `enum` rather than an `int`.

jmarshall (approved):

PR #1354, opened 2025-07-04
[Docs] Visualizing pysam's codebase

ivanmilevtues:
In this PR I am introducing diagram-first documentation. You can see how it renders here:
https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/rexmex/on_boarding.md

The diagrams are designed so that people can quickly grasp the main flow of the project by seeing the main abstract components and how they interact with each other. Since pysam is widely used by scientists and researchers working with genomics data, I believe many are mostly interested in specific parts of the codebase. With these diagrams, they can quickly navigate to that component and start editing it with full context of the pipeline. Further, most researchers focus on the methods rather than the full codebase itself.

Would love to connect and discuss how we can help!

Any feedback is more than welcome! We have introduced a GitHub Action which will keep the docs automatically up to date with new changes, and I’d love to integrate that for you if you’re interested.

Full disclosure: we’re trying to turn this into a startup, but we’re still in a very early stage and figuring out what will actually be useful for people.

### DIFF
--- a/.codeboarding/External_Tool_Wrappers.md
+++ b/.codeboarding/External_Tool_Wrappers.md
@@ -1,0 +1,135 @@
+```mermaid
+
+graph LR
+
+    PysamDispatcher["PysamDispatcher"]
+
+    _pysam_dispatch["_pysam_dispatch"]
+
+    SamtoolsError["SamtoolsError"]
+
+    Samtools_Wrappers["Samtools Wrappers"]
+
+    Bcftools_Wrappers["Bcftools Wrappers"]
+
+    Samtools_Wrappers -- "utilizes" --> PysamDispatcher
+
+    Bcftools_Wrappers -- "utilizes" --> PysamDispatcher
+
+    PysamDispatcher -- "calls" --> _pysam_dispatch
+
+    PysamDispatcher -- "raises" --> SamtoolsError
+
+    _pysam_dispatch -- "can cause" --> SamtoolsError
+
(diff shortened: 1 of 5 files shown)


jmarshall:
There are a number of reasons why this is of no interest to pysam.

> The feedback indicated that source code for pysam.libcalignedsegment.AlignedSegment, pysam.libcbcf.VariantRecord, and pysam.libctabixproxies.NamedTupleProxy could not be retrieved. This is likely due to these being Cython-generated modules,

Pysam is primarily a Cython project, so a tool that cannot handle Cython is of no interest. (This is not an invitation to come back if you do implement Cython support in future.)

> **PysamDispatcher** This component provides the core dispatch mechanism
> **_pysam_dispatch** This is the crucial Foreign Function Interface (FFI) layer, implemented in Cython

In fact these are implementation details that are of no concern to pysam users.

In general both diagrams and text mostly concentrate on implementation trivia rather than the layers that pysam users actually need to know about, and the text is content-free.

> We have introduced a GitHub Action which will keep the docs automatically up to date with new changes, and I’d love to integrate that for you if you’re interested.

We already have detailed documentation that is kept up-to-date in the usual Sphinx and readthedocs way. Rather than helping pysam improve that documentation, you have made a pull request that barely cloaks an advertisement for your vapourware within large amounts of meaningless AI slop text.

[No need to reply. Thanks in advance.]

PR #1358, opened 2025-09-10
Add `VariantRecordSample` documentation

TimD1:
This is a documentation-only PR to introduce documentation for the `VariantRecordSample` class.
- closes #720 
- closes #407

PR #1359, opened 2025-09-11
Add documentation for all `VariantRecord` related classes

TimD1:
This includes the following small classes which were previously missing documentation:
- VariantRecordFilter
- VariantRecordFormat
- VariantRecordInfo
- VariantRecordSamples

Documentation for `VariantRecordSample` was more extensive, and introduced in a previous MR: #1358

### DIFF
--- a/pysam/libcbcf.pyx
+++ b/pysam/libcbcf.pyx
@@ -2238,10 +2238,10 @@ cdef inline int bcf_header_get_info_id(bcf_hdr_t *hdr, key) except? -2:
 ########################################################################
 
 cdef class VariantRecordFilter(object):
-    """Filters set on a :class:`VariantRecord` object, presented as a mapping from
-       filter index or name to :class:`VariantMetadata` object"""
+    """FILTER column data for a :class:`VariantRecord` object, presented as a mapping from
+       a filter index or name (e.g. "PASS") to a :class:`VariantMetadata` object."""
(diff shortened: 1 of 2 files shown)


pysam/libcbcf.pyx
```diff
@@ -2257,7 +2257,7 @@ cdef class VariantRecordFilter(object):
             index = key
 
             if index < 0 or index >= n:
-                raise IndexError('invalid filter index')
+                raise IndexError('Invalid filter index.')
```
jmarshall, line 2260: More than 95% of the error strings in both the Python stdlib and pysam don't have a trailing fullstop, so that's the convention I'd like to follow. I'm on board with capitalising these sentences, but let's leave them without a final period/fullstop.

jmarshall (approved):

jmarshall:
Thanks again for the documentation.

For future reference, note that existing pysam docstrings are written in [NumPy style](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#google-vs-numpy). Yours are mostly in Google style but with NumPy-style type hints (i.e. `arg1 : int` rather than `arg1 (int): blah`) which led to the type hints not being formatted properly by sphinx.

I don't have strong feelings about this (possibly we could use a mixture of styles, with Google style for simpler functions), but for consistency I've largely converted these to the status quo of NumPy style.

PR #1371, opened 2025-12-17
Fix type hints for get_aligned_pairs

gshiba:
Explicitly enumerate all 8 cases. This is a follow up to #1342.

Demo

```python
# test_typing.py
import pysam
a = next(iter(pysam.AlignmentFile("a.bam")))
reveal_type(_ := a.get_aligned_pairs(                                    with_cigar=True ))
reveal_type(_ := a.get_aligned_pairs(                    with_seq=False, with_cigar=True ))
reveal_type(_ := a.get_aligned_pairs(matches_only=False,                 with_cigar=True ))
reveal_type(_ := a.get_aligned_pairs(matches_only=False, with_seq=False, with_cigar=True ))
```

Output: note the difference in inferred type: `List[Tuple[Unknown, ...]]` vs `List[Tuple[int | None, int | None, CIGAR_OPS]]` depending on whether the `False` is explicitly set in the call signature.
```bash
# With mypy
$ uv run --python 3.12 --with pyright --with 'pysam==v0.23.3' --with mypy mypy test_typing.py
test_typing.py:3: note: Revealed type is "builtins.list[builtins.tuple[Any, ...]]"
test_typing.py:4: note: Revealed type is "builtins.list[builtins.tuple[Any, ...]]"
test_typing.py:5: note: Revealed type is "builtins.list[builtins.tuple[Any, ...]]"
test_typing.py:6: note: Revealed type is "builtins.list[tuple[builtins.int | None, builtins.int | None, pysam.libcalignedsegment.CIGAR_OPS]]"
Success: no issues found in 1 source file

# With pyright
$ uv run --python 3.12 --with pyright --with 'pysam==v0.23.3' --with mypy pyright test_typing.py
/home/ec2-user/repos/test_typing.py
  /home/ec2-user/repos/test_typing.py:3:13 - information: Type of "_ := a.get_aligned_pairs(with_cigar=True)" is "List[Tuple[Unknown, ...]]"
  /home/ec2-user/repos/test_typing.py:4:13 - information: Type of "_ := a.get_aligned_pairs(with_seq=False, with_cigar=True)" is "List[Tuple[Unknown, ...]]"
  /home/ec2-user/repos/test_typing.py:5:13 - information: Type of "_ := a.get_aligned_pairs(matches_only=False, with_cigar=True)" is "List[Tuple[Unknown, ...]]"
  /home/ec2-user/repos/test_typing.py:6:13 - information: Type of "_ := a.get_aligned_pairs(matches_only=False, with_seq=False, with_cigar=True)" is "List[Tuple[int | None, int | None, CIGAR_OPS]]"
0 errors, 0 warnings, 4 informations
```


jmarshall:
Thanks for the report and suggested fix. I had tried to ensure the various overloads were disjoint so only one would match any particular invocation, but it's better to cover entirely omitted arguments as well and depend on the ordering of the overloads to ensure invocations get the right one.

Issue #1151, opened 2022-11-19
segfault raised by AlignmentFilePileup_test.py with python3.11


emollier:
Greetings,

With the introduction of python3.11 in Debian sid, Graham Inggs
noticed in [Debian Bug#1024425] that in pysam 0.19.1, the
tests/AlignmentFilePileup_test.py::TestPileupObjects::testIteratorOutOfScope
raises a segmentation fault when running on top of the
python3.11 interpreter.  I could reproduce a similar behavior
while trying to upgrade the pysam package to version 0.20.0.

I attempted to pinpoint the issue by running the faulty test in
pysam 0.20.0 in gdb.  Here is what I could get out of the
debugger:

```
(gdb) run
Starting program: /usr/bin/python3.11 tests/AlignmentFilePileup_test.py
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[Detaching after vfork from child process 1504815]
/build/python-pysam-dNO1D7/python-pysam-0.20.0+ds/tests/PileupTestUtils.py:35: ResourceWarning: unclosed file <_io.TextIOWrapper name='/dev/null' mode='w' encoding='UTF-8'>
  with subprocess.Popen(["samtools", "mpileup", fn],
ResourceWarning: Enable tracemalloc to get the object allocation traceback
.[Detaching after vfork from child process 1504816]
/build/python-pysam-dNO1D7/python-pysam-0.20.0+ds/tests/PileupTestUtils.py:134: ResourceWarning: unclosed file <_io.TextIOWrapper name='/dev/null' mode='w' encoding='UTF-8'>
  return data
ResourceWarning: Enable tracemalloc to get the object allocation traceback
.[Detaching after vfork from child process 1504817]
/build/python-pysam-dNO1D7/python-pysam-0.20.0+ds/tests/PileupTestUtils.py:60: ResourceWarning: unclosed file <_io.TextIOWrapper name='/dev/null' mode='w' encoding='UTF-8'>
  with subprocess.Popen(["samtools", "mpileup", fn] + list(args),
ResourceWarning: Enable tracemalloc to get the object allocation traceback
.[Detaching after vfork from child process 1504818]
/build/python-pysam-dNO1D7/python-pysam-0.20.0+ds/tests/PileupTestUtils.py:60: ResourceWarning: unclosed file <_io.TextIOWrapper name='/dev/null' mode='w' encoding='UTF-8'>
[...]

PR #1380, opened 2026-03-02
Add missing NULL guard in PileupColumn.get_query_qualities

nh13:
## Summary

- Add the missing `plp == NULL` guard to `PileupColumn.get_query_qualities()` to match all 6 sibling methods (`pileups`, `get_num_aligned`, `get_query_sequences`, `get_mapping_qualities`, `get_query_positions`, `get_query_names`)
- Without this guard, calling `get_query_qualities()` after the pileup iterator is exhausted causes a segfault instead of raising a clean `ValueError`

## Test plan

- [ ] Run `REF_PATH=: pytest tests/AlignmentFilePileup_test.py -v`

---

### DIFF
--- a/pysam/libcalignedsegment.pyx
+++ b/pysam/libcalignedsegment.pyx
@@ -3258,6 +3258,9 @@ cdef class PileupColumn:
 
         a list of quality scores : list
         """
+        if self.plp == NULL or self.plp[0] == NULL:
+            raise ValueError("PileupColumn accessed after iterator finished")
+
         cdef uint32_t x = 0
         cdef const bam_pileup1_t * p = NULL
         cdef uint32_t c = 0


nh13:
@jmarshall, apologies for the deluge and the poor timing. I should have reached out first to ask whether these were welcome and when. I've converted all 13 PRs to draft so they're out of your way for the release.

On your points:

1. **Timing**: Totally fair. I'll coordinate better going forward.
2. **AI disclosure**: Done. I've added a disclosure to each PR description. These were AI-assisted using Claude Code: the issues were identified via AI-guided code review, and the implementations were drafted by AI with my review and approval.
3. **Conventional commits**: Done. I've rewritten all commit messages and PR titles to match the project's style.

Happy to close any or all of these if they're not useful, or leave them as drafts until you're open to reviewing them. Whatever works best for you.

@bioinformed, thanks for the kind words and the offer to help review.

jmarshall:
There are test cases for these accessors — see `testIteratorOutOfScope()` in _tests/AlignmentFilePileup_test.py_. On older Pythons, they all work as expected with or without this PR — in the case of `get_query_qualities()` a ValueError is raised in these circumstances a few lines lower down. Nonetheless it would be good to apply your PR to regularise the validity checks.

However none of this works in Python 3.11, which is #1151. Rather to my surprise, it turns out these test cases do work again — and don't crash — in Python 3.12 and later! Nonetheless I'd like to use this PR as an additional excuse to get to the bottom of that Python 3.11 problem and hopefully merge a fix for that at the same time.

PR #1381, opened 2026-03-02
Fix TypeError in modified_bases_forward when query_sequence is None

nh13:
## Summary

- Fix a `TypeError` in the `modified_bases_forward` property when `query_sequence` is `None` but `infer_read_length()` returns a value
- The previous logic checked `rlen is None and self.query_sequence is None`, which meant when `infer_read_length()` returned a non-None value but `query_sequence` was still `None`, the code fell through to `len(self.query_sequence)` causing a `TypeError`
- The `infer_read_length()` result was never actually used — only `query_sequence` is needed for the coordinate transform, so the guard is simplified to check only `query_sequence`

## Test plan

- [ ] Run `REF_PATH=: pytest tests/AlignedSegment_test.py -v`

---

### DIFF
--- a/pysam/libcalignedsegment.pyx
+++ b/pysam/libcalignedsegment.pyx
@@ -1939,12 +1939,9 @@ cdef class AlignedSegment:
             if pmods and self.is_reverse:                
                 rmod = {}
 
-                # Try to find the length of the original sequence
-                rlen = self.infer_read_length()
-                if rlen is None and self.query_sequence is None:
+                if self.query_sequence is None:
                     return rmod
-                else:
-                    rlen = len(self.query_sequence)
+                rlen = len(self.query_sequence)
                     
                 for k,mods in pmods.items():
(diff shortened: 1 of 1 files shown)


jmarshall:
The code is redundant, but your cybernetic friend's analysis is incomplete.

The SEQ field is also used by the C code underlying `modified_bases`. In particular, if `query_sequence` is None then `bam_next_basemod()` will (usually print a diagnostic and) always return 0, and hence `modified_bases` and `pmods` will be `{}`. Or possibly `bam_next_basemod()` might return -1 signalling error, in which case `pmods` will be None. Either way, consequently when `query_sequence` is None the code in question will never be executed, so no TypeError can be raised here.

What `modified_bases_forward` needs is a consideration of what its semantics ought to be, and for it to be rewritten accordingly. For example, it should probably reverse each `mods` list so that the coordinates remain in ascending order.

PR #1385, opened 2026-03-02
Remove redundant copy in getSequenceInRange

nh13:
## Summary

- Remove a redundant copy in `getSequenceInRange` where `charptr_to_bytes(seq)` was called on `seq` which is already a Python `bytes` object created by `PyBytes_FromStringAndSize`
- The `charptr_to_bytes` call was converting `bytes → const char* → bytes`, creating an unnecessary copy

## Test plan

- [ ] Run `REF_PATH=: pytest tests/AlignedSegment_test.py -v`

---


jmarshall:
Thanks. Removing this silly mistake does make an observable difference to the speed.

However we can do better — now also superseded by PR #1411.

PR #1392, opened 2026-03-02
Extract shared PileupColumn validity check

nh13:
## Summary

- Extract a `_check_pileup_valid()` cdef helper on `PileupColumn` that raises `ValueError` if the pileup buffer is NULL (iterator exhausted)
- All 7 public PileupColumn methods now call this helper instead of inlining the guard check
- Also adds the previously missing guard to `get_query_qualities()` (fixing the same bug as PR #1380 but structurally)

## Test plan

- [ ] Run `REF_PATH=: pytest tests/AlignmentFilePileup_test.py -v`

---

### DIFF
--- a/pysam/libcalignedsegment.pxd
+++ b/pysam/libcalignedsegment.pxd
@@ -85,6 +85,7 @@ cdef class PileupColumn:
     cdef uint32_t min_base_quality
     cdef kstring_t buf
     cdef char * reference_sequence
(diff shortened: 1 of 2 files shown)


jmarshall:
It's a minor thing, but extracting this into a helper function will add another entry to the stack backtrace. So I'd prefer to apply your #1380 — however this is worse than it looks; see the comment there.

PR #1406, opened 2026-04-28
Accept file-like objects in `VariantFile` filename parameter

dpoznik:
The `VariantFile.__init__` and `VariantFile.open` stubs typed `filename` as `StrOrBytesPath`. This excluded file-like objects like `TextIO` and `BinaryIO`, but the underlying implementation supports them. When `filename` is not a path, `open()` catches the `TypeError` from `encode_filename`, then `_open_htsfile()` calls `.fileno()` on it to obtain a file descriptor that is passed to htslib's `hdopen`/`hts_hopen`. Consequently, any object with a `.fileno()` method is valid.

### DIFF
--- a/pysam/libcbcf.pyi
+++ b/pysam/libcbcf.pyi
@@ -1,8 +1,8 @@
-import sys
 from typing import (
     Optional,
     Union,
     Any,
+    IO,
     Sequence,
     Tuple,
     Iterator,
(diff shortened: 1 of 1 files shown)


jmarshall:
Well, that's annoying. I had traced through the code of `AlignmentFile` and `VariantFile`. `AlignmentFile` has relatively obvious code that handles all these alternatives; but, as you say, it turns out that `VariantFile` has subtler code that effectively does the same thing too. So I had wrongly concluded that `VariantFile` did not accept these.

Thanks for the report. I've fixed it in a different way, and added some type checking tests.

dpoznik:
Cool, thanks for the fix!

